PR #10, opened 2020-11-03
Add decriptions of polynomial set and basis and dual set

mscroggs:


chrisrichardson (approved):
Let's merge this. We can do more later.

PR #11, opened 2020-11-03
Use scikit-build to build and install with pip

chrisrichardson:
This builds and installs using scikit-build.
Key points:
 * makes `libtab` "pip-installable" - `pip3 install --user .` will work, for example.
 * Can also use `cmake` to build the project, if desired (with `-DCMAKE_INSTALL_PREFIX` etc.)
 * Can also use `python3 setup.py install` if desired.
 * Installs both `libtab.so` and `libtab.cpython.foo.so` (pybind) to the same location, and sets RPATH to fix linking of wrapper code.
 * So far, does not install any C++ headers or C/C++ API, but this could be added, once it is decided on the structure.


chrisrichardson:
Reduced pre-requisites down to just having Eigen and scikit-build installed, and testing on MacOS.

PR #17, opened 2020-11-06
Add DOF permutations and direction correction

mscroggs:
I've added DOF permutations (#6) for Lagrange and Nedelec (first kind) elements, and vector direction correction for Nedelec (first kind).

We maybe don't want to merge this until I've added these for the other element, but we should think about whether this is the right structure before I do more elements.

I still need to make finish writing tests and/or make some better tests that this is correct.


chrisrichardson:
Eigen also has a permutation matrix class, I don't know if that would be useful here.
https://eigen.tuxfamily.org/dox/classEigen_1_1PermutationMatrix.html

cpp/wrapper.cpp
```diff
@@ -73,16 +73,24 @@ Each element has a `tabulate` function which returns the basis functions and a n
...
-        return FiniteElement(celltype, degree, value_size, new_coeffs);
+        return FiniteElement(celltype, degree, value_size, new_coeffs,
+                             base_permutations);
       },
       "Create an element from basic data");
 
   py::class_<FiniteElement>(m, "FiniteElement", "Finite Element")
       .def("tabulate", &FiniteElement::tabulate, tabdoc.c_str())
+      .def("base_permutations", &FiniteElement::base_permutations,
+           tabdoc.c_str())
```
chrisrichardson, line 90: remove `tabdoc.c_str()`

PR #55, opened 2020-11-17
Eigen type updates and simplifications

garth-wells:


cpp/lagrange.cpp
```diff
@@ -18,21 +19,19 @@ FiniteElement Lagrange::create(cell::Type celltype, int degree)
...
+  // Create points at nodes, ordered by topology (vertices first)
+  Eigen::ArrayXXd pt(ndofs, topology.size() - 1);
   if (degree == 0)
   {
     pt = lattice::create(celltype, 0, lattice::Type::equispaced, true);
-    for (int i = 0; i < tdim; ++i)
+    for (std::size_t i = 0; i < entity_dofs.size(); ++i)
       entity_dofs[i].resize(topology[i].size(), 0);
-    entity_dofs[tdim] = {1};
+    entity_dofs[topology.size() - 1][0] = 1;
```
chrisrichardson, line 34: I preferred how it was before. Couldn't we have `const int tdim = topology.size() - 1;` at line 27?

garth-wells: We could have `const int tdim = topology.size() - 1;`. What was there before was not good because the loop range (`tdim`) was computed completely independently of `topology`, which is confusing and error prone.

PR #62, opened 2020-11-19
Added family name string to finite elments

mscroggs:
Store the family name so this can be put in the FFC generated code then used by dolfin to reconstruct the element


chrisrichardson:
Could we make the family name as the first argument to FiniteElement?

chrisrichardson:
How can we match this up with the code in `util.cpp` - we don't want to have the names appear twice.

mscroggs:
> How can we match this up with the code in `util.cpp` - we don't want to have the names appear twice.

I found a way in 88d8c56

PR #67, opened 2020-11-20
More refactoring towards a public API

garth-wells:


cpp/libtab.cpp
```diff
@@ -4,33 +4,155 @@
...
-                    {dlagrange::family_name, &dlagrange::create},
-                    {lagrange::family_name, &lagrange::create},
-                    {nedelec::family_name, &nedelec::create},
-                    {nedelec2::family_name, &nedelec2::create},
-                    {rt::family_name, &rt::create},
-                    {regge::family_name, &regge::create}};
-
-  auto create_it = create_map.find(family);
-  if (create_it == create_map.end())
+  if (family == "Lagrange")
```
chrisrichardson, line 23: I find the const map neater than if/elseif

cpp/libtab.h
```diff
@@ -1,25 +1,340 @@
...
+  /// Get the element value size
+  /// This is just a convenience function returning product(value_shape)
+  /// @return Value size
+  int value_size() const;
+
+  /// Get the element value tensor shape, e.g. returning [1] for scalars.
+  /// @return Value shape
+  const std::vector<int>& value_shape() const;
+
+  /// Dimension of the finite elemeent space (number of degrees of
```
chrisrichardson, line 76: elemeent

cpp/libtab.cpp
garth-wells: The mucking about with maps to function pointers is a bit ugly.

chrisrichardson: `else if` is ugly. matter of opinion.

Issue #15, opened 2020-11-05
Add interface for performing interpolation


garth-wells:
It would be helpful to support interpolation. 
- For an element, return the points on the reference cell where it requires function evaluations
- Given function evaluations on the reference element, return dofs.

chrisrichardson:
This is trivial, I guess for Lagrange. For other elements, in principle, we should be able to just have the same number of sample points as dofs, and then just do a matrix inversion/solve.

Questions to think about
-------------
* The set of points is non-unique - how should we choose them?
* Should we support oversampling... and least squares fitting?
* How does this relate to the mesh? Are we considering only interpolation on the same cell, or non-matching meshes (e.g. for mesh refinement). Interpolation from DG to CG spaces?

PR #77, opened 2020-12-02
Interpolation data

mscroggs:
Added a function that return a matrix of weights and array of points to represent how to interpolate into a finite element (#15).

For example:

- for Lagrange spaces, the points will be the dof locations, and the matrix is the identity.
- for spaces with integral moments, the points will the quadrature points, and the matrix will contain the quadrature weights.

Currently this is only implemented for Lagrange spaces.

### DIFF
--- a/cpp/lagrange.cpp
+++ b/cpp/lagrange.cpp
@@ -125,15 +125,8 @@ FiniteElement libtab::create_lagrange(cell::type celltype, int degree,
     }
   }
 
-  std::vector<std::pair<Eigen::ArrayXd, Eigen::ArrayXXd>> interpolation_info;
-  for (int i = 0; i < pt.rows(); ++i)
-  {
-    Eigen::ArrayXd a(1);
-    a(0) = 1;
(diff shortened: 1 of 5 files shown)


chrisrichardson:
This doesn't add any new information (yet), that isn't already in `_points`. At least, we need to remove `_points` if we are going to have `_interpolation_info` instead. "info" is also not a great name. 

So, I'd suggest, either: 
1. remove `_points` and rename `_interpolation_info` to something better e.g. `_interpolation_rule` ?, or
2. keep `_points` and and replace `_interpolation_info` with `Eigen::MatrixXd _interpolation_matrix`... 

maybe?

mscroggs:
I've gone for (2). Seems most sensible as people may find a use for the points or matrix without needing the other at the same time

chrisrichardson:
Minor point, but they would look better grouped together in the file, rather than with `_family` somewhat incongruously interjected.

mscroggs:
I moved family name up as it made sense to put it with the cell type and order

PR #86, opened 2020-12-22
Implement permutations for Nedelec second kind on tetrahedra

mscroggs:


chrisrichardson:
Seems fine. Just merge if you're happy with it.

Issue #97, opened 2021-01-18
Expose cell::str_to_type and cell::type_to_str to python layer


jorgensd:
To make it as easy as possible to use basix along-side dolfin-x, I suggest that `cell:str_to_type` and `cell::type_to_str` is expose, such that one can easily convert between `dolfinx::mesh::CellType` and `basix::CellType`.
This is for instance relevant when creating custom quadrature rules, as `make_quadrature` require a `basix` cell type. 
In dolfin-x, one can convert the `dolfinx::mesh::CellType` to a string using `dolfinx.cpp.mesh.to_string(mesh.topology.cell_type)`

PR #107, opened 2021-01-29
Add mappings

### LINKED ISSUE #99: Implement Piola mappings
These can then be used directly by dolfinx (see FEniCS/ffcx#299)

### LINKED ISSUE #102: Fix mapping name
Should be "identity" as per UFL, not "affine".

mscroggs:
- Added `mapping::type` enum to use instead of string identifier.
- Implemented Piola mappings.
- Will allow us to remove `transform_reference_basis` from ffc and use these mappings directly.

Fixes #102. Fixes #99.

Progress towards #97 (type_to_str is now wrapped)

This PR is coupled with FEniCS/ffcx#304 and FEniCS/dolfinx#1369

### DIFF
--- a/cpp/basix.h
+++ b/cpp/basix.h
@@ -35,7 +35,7 @@ int dim(int handle);
 /// Family name
 const char* family_name(int handle);
 
-/// Mapping name (affine, piola etc.)
+/// Mapping name (identity, piola etc.)
 const char* mapping_name(int handle);
 
 /// Number of dofs per entity, ordered from vertex, edge, facet, cell
--- a/cpp/basix.h
+++ b/cpp/basix.h
@@ -17,6 +17,11 @@ void release_element(int handle);
 std::vector<Eigen::ArrayXXd> tabulate(int handle, int nd,
                                       const Eigen::ArrayXXd& x);
 
+/// Apply mapping
+Eigen::ArrayXXd apply_mapping(int handle, const Eigen::ArrayXd& reference_data,
+                              const Eigen::MatrixXd& J, double detJ,
+                              const Eigen::MatrixXd& K);
+
 /// Cell type
 const char* cell_type(int handle);
 
--- a/cpp/basix.h
+++ b/cpp/basix.h
@@ -17,11 +17,32 @@ void release_element(int handle);
 std::vector<Eigen::ArrayXXd> tabulate(int handle, int nd,
                                       const Eigen::ArrayXXd& x);
 
-/// Apply mapping to push reference data to the physical cell
+/// Push reference data forward to the physical cell
+/// @param[in] handle The handle of the basix element
+/// @param[in] reference_data The reference data at a single point
+/// @param[in] J The Jacobian of the map to the cell (evaluated at the point)
+/// @param[in] detJ The determinant of the Jacobian of the map to the cell
+/// (evaluated at the point)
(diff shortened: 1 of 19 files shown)


cpp/basix.h
```diff
@@ -17,6 +17,11 @@ void release_element(int handle);
 std::vector<Eigen::ArrayXXd> tabulate(int handle, int nd,
                                       const Eigen::ArrayXXd& x);
 
+/// Apply mapping
```
garth-wells, line 20: What map? I could suggest a name but I don't know from the docstring what it does. How many point does it work on? Can the operation it performs be described?

cpp/finite-element.h
```diff
@@ -215,9 +216,29 @@ class FiniteElement
...
-  const std::string& mapping_name() const;
+  const mapping::type mapping_type() const;
+
+  /// Map a function value from the reference to a cell
+  /// @param reference_data The function value on the reference
+  /// @param J The Jacobian of the mapping
+  /// @param detJ The determinant of the Jacobian of the mapping
+  /// @param K The inverse of the Jacobian of the mapping
+  /// @return The function value on the cell
+  Eigen::ArrayXd apply_mapping(const Eigen::ArrayXd& reference_data,
```
garth-wells, line 229: Could we call this `map_push_forward`?

cpp/finite-element.h
```diff
@@ -215,9 +216,29 @@ class FiniteElement
...
+                               const Eigen::MatrixXd& J, double detJ,
+                               const Eigen::MatrixXd& K) const;
+
+  /// Map a function value from a cell to the reference
+  /// @param physical_data The function value on the cell
+  /// @param J The Jacobian of the mapping
+  /// @param detJ The determinant of the Jacobian of the mapping
+  /// @param K The inverse of the Jacobian of the mapping
+  /// @return The function value on the reference
+  Eigen::ArrayXd apply_inverse_mapping(const Eigen::ArrayXd& physical_data,
```
garth-wells, line 239: Could we call this `map_pull_back`?

cpp/finite-element.h
```diff
@@ -215,9 +216,29 @@ class FiniteElement
   /// @return The family name
   const std::string& family_name() const;
 
-  /// Get the mapping used for this element
+  /// Get the mapping type used for this element
   /// @return The mapping
-  const std::string& mapping_name() const;
+  const mapping::type mapping_type() const;
+
+  /// Map a function value from the reference to a cell
```
garth-wells, line 223: Should this be in the public API header file? (`basix.h`)?

```
/// Map a function value on the reference cell to the physical cell
```
?

garth-wells (changes requested):
I've added a few comments on clarifications. Can look more closely once a few points are clarified.

cpp/basix.h
mscroggs: I've renamed this to map_push_forward and added a better docstring.

cpp/finite-element.h
mscroggs: It's now in both

chrisrichardson (approved):

mscroggs:
FEniCS/dolfinx#1369 needs a bit of attention still, but I can make a minor change in ffcx to allow this one to be merged without breaking dolfin

PR #113, opened 2021-02-05
[WIP] Template mappings

mscroggs:
- Added a template parameter to the mappings for the value type
- Added real and complex variants of the `map_push_forward` and `map_pull_back` functions to basix.h


chrisrichardson:
This might need rethinking for a C-style interface, so maybe we can hold off until that is done.

PR #114, opened 2021-02-08
Avoid dividing by zero

massimiliano-leoni:
`h` was 1 / 0 for DG0 elements. Although `h` was never actually used when `n == 0`, this triggered some annoying floating-point exception traps while debugging.


mscroggs (approved):

chrisrichardson:
I'd rather just put `h` inside the `case` for each cell type.

massimiliano-leoni:
@chrisrichardson I don't mind either way.

chrisrichardson:
@massimiliano-leoni - can you try it the other way? I think `h` is only used by interval, quadrilateral and hexahedron anyway.

massimiliano-leoni:
@chrisrichardson there you have it!

mscroggs (approved):

chrisrichardson (approved):

PR #115, opened 2021-02-09
Enhance cmake installation mode to install python files

draenog:


garth-wells:
Install process has been completely overhauled, so this it out-of-date now.

PR #118, opened 2021-02-11
Remove Eigen from basix.h

### LINKED ISSUE #96: Remove EIgen from public interface and use plain C API
Exposing Eigen in the public interface can cause all sort of problems when interfacing with other libraries that may use a different alignment. Simple fix is to not expose Eigen.

mscroggs:
Fixes #96.

### DIFF
--- a/cpp/basix.cpp
+++ b/cpp/basix.cpp
@@ -125,9 +125,10 @@ void basix::interpolation_points(int handle, double* points)
 void basix::interpolation_matrix(int handle, double* matrix)
 {
   check_handle(handle);
-  Eigen::Map<Eigen::MatrixXd>(matrix, dim(handle),
-                              interpolation_num_points(handle)
-                                  * _registry[handle]->value_size())
+  Eigen::Map<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>(matrix,
+                              dim(handle),
+                              interpolation_num_points(handle) * _registry[handle]->value_size()
+                              )
       = _registry[handle]->interpolation_matrix();
 }
 
--- a/cpp/basix.cpp
+++ b/cpp/basix.cpp
@@ -2,6 +2,7 @@
 #include "basix.h"
 #include "cell.h"
 #include "finite-element.h"
+#include "quadrature.h"
 #include <memory>
 #include <vector>
 
@@ -105,4 +106,11 @@ basix::topology(const char* cell_type)
 {
   cell::type ct = cell::str_to_type(cell_type);
   return cell::topology(ct);
+}
+
+std::pair<Eigen::ArrayXXd, Eigen::ArrayXd>
+basix::make_quadrature(const char* rule, const char* cell_type, int order)
+{
+  cell::type ct = cell::str_to_type(cell_type);
+  return basix::quadrature::make_quadrature(rule, ct, order);
 }
--- a/cpp/basix.cpp
+++ b/cpp/basix.cpp
@@ -85,6 +85,12 @@ int basix::degree(int handle)
   return _registry[handle]->degree();
 }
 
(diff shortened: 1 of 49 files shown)


cpp/basix.cpp
```diff
@@ -117,22 +179,28 @@ const char* basix::mapping_name(int handle)
...
-std::vector<std::vector<std::vector<int>>>
-basix::topology(const char* cell_type)
+int basix::cell_geometry_dimension(const char* cell_type)
 {
   cell::type ct = cell::str_to_type(cell_type);
-  return cell::topology(ct);
+  return cell::geometry(ct).cols();
+}
+
+void basix::cell_geometry(const char* cell_type, double* points)
```
garth-wells, line 194: Does this function get called frequently? (had a dynamic memory allocation in it)

cpp/finite-element.cpp
```diff
@@ -159,6 +156,30 @@ const std::vector<std::vector<int>>& FiniteElement::entity_dofs() const
...
+  int ndsize = 1;
+  for (int i = 1; i <= nd; ++i)
+    ndsize *= (tdim + i);
+  for (int i = 1; i <= nd; ++i)
+    ndsize /= i;
+
+  const int ndofs = _coeffs.rows();
+  const int vs = value_size();
+
+  std::vector<double> basis_data(ndsize * x.rows() * ndofs * vs);
```
garth-wells, line 170: Can we avoid this dynamic allocation?

cpp/finite-element.cpp
```diff
@@ -170,44 +191,51 @@ FiniteElement::tabulate(int nd, const Eigen::ArrayXXd& x) const
...
   return _base_permutations;
 }
 //-----------------------------------------------------------------------------
+int FiniteElement::num_points() const { return _points.rows(); }
+//-----------------------------------------------------------------------------
 const Eigen::ArrayXXd& FiniteElement::points() const { return _points; }
 //-----------------------------------------------------------------------------
-Eigen::ArrayXd
-FiniteElement::map_push_forward(const Eigen::ArrayXd& reference_data,
+Eigen::ArrayXXd
```
garth-wells, line 217: Should this be an in-out argument?

cpp/finite-element.cpp
```diff
@@ -170,44 +191,51 @@ FiniteElement::tabulate(int nd, const Eigen::ArrayXXd& x) const
...
+    result.col(i) = mapping::map_push_forward(reference_data.col(i), J, detJ, K,
+                                              _mapping_type, _value_shape);
+  return result;
 }
 //-----------------------------------------------------------------------------
-Eigen::ArrayXd FiniteElement::map_pull_back(const Eigen::ArrayXd& physical_data,
-                                            const Eigen::MatrixXd& J,
-                                            double detJ,
-                                            const Eigen::MatrixXd& K) const
+Eigen::ArrayXXd
```
garth-wells, line 229: Out argument?

cpp/finite-element.cpp
mscroggs: Yes, I've added this to #119

cpp/finite-element.cpp
mscroggs: Yes, I've added this to #119

cpp/finite-element.cpp
mscroggs: This version of tabulate is only used by the Python wrapper. The c-interface uses `tabulate_to_memory`

cpp/basix.cpp
mscroggs: No, this was changed to get rid of Eigen. I believe dolfinx will call this once then store the points

chrisrichardson (approved):

garth-wells (approved):

PR #135, opened 2021-02-25
Add span and array2d, start some transitions

garth-wells:


mscroggs:
`dolfinx/python/test/unit/fem/test_custom_jit_kernels.py::test_coefficient` is giving double free or corruption when I use this branch

mscroggs:
Sorry, scratch that. I'm getting the same error on master. Something wrong with my docker image I think...

mscroggs (approved):

Issue #131, opened 2021-02-25
Rename base permutations to base transformations


mscroggs:

garth-wells:
Can this be closed now?

PR #140, opened 2021-03-04
Add Serendipity element

mscroggs:
- Added arbitrary order serendipity element on interval, quad and hex.
- Started using interpolation to calculate the base transformations for integral moment dofs, and removed a lot of the old dof transformation code.
- Renamed permutations -> transformations internally (#131). This still need updating in the public interface and in DOLFINx/FFCx


mscroggs:
I need to replace to TODOs in moments.h before we can merge this

garth-wells (approved):

PR #161, opened 2021-03-15
Wrong tensor size in quadrature

mscroggs:
In `quadrature::compute_gauss_jacobi_points`, the number of dimensions of a tensor was incorrect. A few other changes:

- ~Moved the -O2 to all builds, so that Release is sped up.~
- Added `-DCMAKE_BUILD_TYPE=Release` to the install instructions


.github/workflows/pythonapp.yml
```diff
@@ -72,3 +72,34 @@ jobs:
           git add --all
           git commit --allow-empty -m "C++ FEniCS/basix@${{ github.sha }}"
           git push
+
+  build-not-dev:
```
garth-wells, line 76: What flags are different for this build case? Ideally we'd pick up these issues on a single build and run.

mscroggs: I doesn't have `-DCMAKE_BUILD_TYPE=Developer`. For reasons beyond me some errors were only happening when building without developer mode (see eg https://github.com/bempp/bempp-cl/runs/2110771847)

garth-wells: The CMake build type sets the flags - need to understand which compiler flags need to be set.

chrisrichardson:
Can we merge this, to fix the stack smashing?

mscroggs:
Seems sensible. I don't think it will cause any conflicts with the xtensor branch

garth-wells:
The changes to the CMake file are confusing to me - the logic is hard to follow. Seems wrong to force the optimisation level in debug mode.

mscroggs:
Agreed. I think it makes sense to have `-O2` on Release and Developer but not Debug. Just pushed a change to do this

PR #179, opened 2021-04-01
Hotfix - xtensor 0.23.4 not compatible with 0.23.2

IgorBaratta:
It breaks basix if xtensor is not updated.
This PR introduces xtensor and xtl minimum versions.


garth-wells (approved):

chrisrichardson:
This is a bit annoying, as the spack PR just got approved with the previous 0.23.2 version

garth-wells:
xtensor folk fixed some bugs that we found.

Issue #100, opened 2021-01-19
Implement combining of base transformations


mscroggs:
It would be useful to have a function that takes a `uint32_t` that decodes the cell orientation, then returns the permutation. Dolfinx could then use this instead of the version in the ffc generated code (see FEniCS/ffcx#300)

PR #184, opened 2021-04-07
Reduce amount stored for DOF transformations

mscroggs:
Instead of storing 1 or 2 `ndofs` by `ndofs` matrices for each edge and face, we now store 1 or 2 `num_entity_dofs` by `num_entity_dofs` matrices for each subentity type.

eg for an order 4 Lagrange space on a hexahedron, we used to store 24 125x125 matrices, but we now store one 3x3 matrix and two 9x9 matrices.

This has simplified the generation of base transformations for each element.

This is progress towards #100.

This PR is coupled with FEniCS/ffcx#322

### DIFF
--- a/cpp/basix/lagrange.cpp
+++ b/cpp/basix/lagrange.cpp
@@ -103,128 +103,61 @@ FiniteElement basix::create_lagrange(cell::type celltype, int degree)
     }
   }
 
-  std::size_t transform_count = 0;
-  for (std::size_t i = 1; i < topology.size() - 1; ++i)
-    transform_count += topology[i].size() * i;
-  auto base_transformations
-      = xt::tile(xt::expand_dims(xt::eye<double>(ndofs), 0), transform_count);
-  switch (celltype)
-  {
-  case cell::type::interval:
-    assert(transform_count == 0);
-    break;
-  case cell::type::triangle:
+  std::vector<xt::xtensor<double, 2>> entity_transformations;
+  if (tdim > 1)
   {
     const std::vector<int> edge_ref
         = doftransforms::interval_reflection(degree - 1);
-    for (int edge = 0; edge < 3; ++edge)
-    {
-      auto bt = xt::view(base_transformations, edge, xt::all(), xt::all());
-      const int start = 3 + edge_ref.size() * edge;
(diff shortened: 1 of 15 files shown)


cpp/basix/lagrange.cpp
```diff
@@ -103,128 +103,61 @@ FiniteElement basix::create_lagrange(cell::type celltype, int degree)
...
-      for (std::size_t i = 0; i < edge_ref.size(); ++i)
-      {
-        bt(start + i, start + i) = 0;
-        bt(start + i, start + edge_ref[i]) = 1;
-      }
-    }
-    break;
+    xt::xtensor<double, 2> et
+        = xt::zeros<double>({edge_ref.size(), edge_ref.size()});
+    for (std::size_t i = 0; i < edge_ref.size(); ++i)
```
garth-wells, line 113: I think something like
```cpp
xt::view(et, xt::all(), xt::keep(edge_ref)) = 1;
```
should work here.

cpp/basix/lagrange.cpp
```diff
@@ -103,128 +103,61 @@ FiniteElement basix::create_lagrange(cell::type celltype, int degree)
...
-                          xt::all());
-      const int start = 4 + edge_ref.size() * 6 + face_ref.size() * face;
-      for (std::size_t i = 0; i < face_rot.size(); ++i)
-      {
-        bt0(start + i, start + i) = 0;
-        bt0(start + i, start + face_rot[i]) = 1;
-        bt1(start + i, start + i) = 0;
-        bt1(start + i, start + face_ref[i]) = 1;
-      }
+      ft0(i, face_rot[i]) = 1;
```
garth-wells, line 129: See above on possible use of `xt::view`.

cpp/basix/lagrange.cpp
mscroggs: amazing

mscroggs: `xt::view(et, xt::all(), xt::keep(edge_ref))` gets the full block, not just the few entries we want. Can't immediately see a way to do this with a view

garth-wells (approved):

PR #187, opened 2021-04-08
Add permute_dofs and unpermute_dofs functions to Basix

mscroggs:
Added permute_dofs and unpermute_dofs functions to Basix, so these can be removed from the generated code.

Progress towards #100.

This PR is coupled with FEniCS/ffcx#325 and FEniCS/dolfinx#1471.

### DIFF
--- a/cpp/basix/finite-element.cpp
+++ b/cpp/basix/finite-element.cpp
@@ -358,6 +358,7 @@ FiniteElement::FiniteElement(
     for (std::size_t i = 0; i < _entity_transformations.size(); ++i)
     {
       std::vector<int> perm(_entity_transformations[i].shape(0));
+      std::vector<int> rev_perm(_entity_transformations[i].shape(0));
       for (std::size_t row = 0; row < _entity_transformations[i].shape(0);
            ++row)
       {
@@ -367,11 +368,31 @@ FiniteElement::FiniteElement(
           if (_entity_transformations[i](row, col) > 0.5)
           {
             perm[row] = col;
+            rev_perm[col] = row;
             break;
           }
         }
       }
-      _entity_permutations.push_back(perm);
+      // Factorise the permutations
+      std::vector<int> f_perm(perm.size());
+      for (std::size_t row = 0; row < perm.size(); ++row)
+      {
+        std::size_t row2 = perm[row];
+        while (row2 < row)
+          row2 = perm[row2];
+        f_perm[row] = row2;
+      }
+      _entity_permutations.push_back(f_perm);
+
+      std::vector<int> f_rev_perm(rev_perm.size());
+      for (std::size_t row = 0; row < rev_perm.size(); ++row)
+      {
+        std::size_t row2 = rev_perm[row];
+        while (row2 < row)
+          row2 = rev_perm[row2];
+        f_rev_perm[row] = row2;
+      }
+      _reverse_entity_permutations.push_back(f_rev_perm);
     }
   }
 }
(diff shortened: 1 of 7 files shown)


cpp/basix/finite-element.cpp
```diff
@@ -335,6 +335,78 @@ FiniteElement::FiniteElement(
...
+  {
+    for (std::size_t row = 0; row < _entity_transformations[i].shape(0); ++row)
+    {
+      double rmin
+          = xt::amin(xt::view(_entity_transformations[i], row, xt::all()))(0);
+      double rmax
+          = xt::amax(xt::view(_entity_transformations[i], row, xt::all()))(0);
+      double rtot
+          = xt::sum(xt::view(_entity_transformations[i], row, xt::all()))(0);
+      if ((_entity_transformations[i].shape(1) != 1
```
garth-wells, line 354: These values look a bit arbitrary. Is these a neater way?

cpp/basix/finite-element.cpp
```diff
@@ -533,6 +615,140 @@ xt::xtensor<double, 3> FiniteElement::map_pull_back(
...
+void FiniteElement::permute_dofs(tcb::span<int>& dofs,
+                                 std::uint32_t cell_info) const
+{
+  if (!_dof_transformations_are_permutations)
+    throw std::runtime_error(
+        "The DOF transformations for this element are not permutations");
+  if (_dof_transformations_are_identity)
+    return;
+
+  const int tdim = cell::topological_dimension(_cell_type);
```
garth-wells, line 627: Should the `FiniteElement` object store the `tdim` to avoid calling this function a lot?

cpp/basix/finite-element.cpp
```diff
@@ -533,6 +615,140 @@ xt::xtensor<double, 3> FiniteElement::map_pull_back(
...
+void FiniteElement::unpermute_dofs(tcb::span<int>& dofs,
+                                   std::uint32_t cell_info) const
+{
+  if (!_dof_transformations_are_permutations)
+    throw std::runtime_error(
+        "The DOF transformations for this element are not permutations");
+  if (_dof_transformations_are_identity)
+    return;
+
+  const int tdim = cell::topological_dimension(_cell_type);
```
garth-wells, line 694: See above comment.

cpp/basix/finite-element.cpp
```diff
@@ -533,6 +615,140 @@ xt::xtensor<double, 3> FiniteElement::map_pull_back(
...
+  if (tdim >= 2)
+  {
+    // This assumes 3 bits are used per face. This will need updating if 3D
+    // cells with faces with more than 4 sides are implemented
+    int face_start = tdim == 3 ? 3 * cell::sub_entity_count(_cell_type, 2) : 0;
+
+    int dofstart = 0;
+    for (int v = 0; v < cell::sub_entity_count(_cell_type, 0); ++v)
```
garth-wells, line 702: Should `cell::sub_entity_count` be stored bu the `FiniteElement`?

cpp/basix/finite-element.h
```diff
@@ -421,6 +429,16 @@ class FiniteElement
   /// ~~~~~~~~~~~~~~~~
   xt::xtensor<double, 3> base_transformations() const;
 
+  /// Permute the dof numbering on a cell
+  /// @param[in,out] dofs The dof numbering for the cell
+  /// @param cell_info The permutation info for the cell
+  void permute_dofs(tcb::span<int>& dofs, std::uint32_t cell_info) const;
```
garth-wells, line 435: Perhaps the span should be for a fixed-width integer type?

garth-wells (changes requested):
Few small suggestions.

cpp/basix/finite-element.cpp
mscroggs: I've made them use an `isClose` function to make it more readable but can't find a better way to compare two floats. For all the elements so far the nearest we get to 1 or 0 without being equal is 0.5, so the large epsilon comparison is good enough for now

garth-wells: `xt::close` [1] not suitable?

[1] https://xtensor.readthedocs.io/en/latest/api/classif_functions.html#_CPPv4I00EN2xt7iscloseEDaRR2E1RR2E2ddb

PR #195, opened 2021-04-18
Remove c interface

garth-wells:
No longer used by DOLFINx, and added quite a bit of complexity.

Can always return in future if there is demand.


mscroggs:
Nice

mscroggs (approved):

PR #197, opened 2021-04-20
A Numba dof transformations function

### LINKED ISSUE #97: Expose cell::str_to_type and cell::type_to_str to python layer

mscroggs:
This helper function will allow people writing custom assemblers in DOLFINx to use dof transformations without having to implement from scratch.

Fixes #97.

### DIFF
--- a/python/basix/numba_helpers.py
+++ b/python/basix/numba_helpers.py
@@ -1,0 +1,56 @@
+try:
+    import numba
+except ImportError:
+    raise RuntimeError("You must have numba installed to use the numba helper functions.")
+
+
+@numba.njit
+def apply_dof_transformation(entity_transformations, cell_type, entity_dofs,
+                             data, block_size, cell_info):
+    if cell_type == "interval":
+        tdim = 1
+    elif cell_type == "triangle":
+        edge_count = 3
+        tdim = 2
+    elif cell_type == "quadrilateral":
+        edge_count = 4
+        tdim = 2
+    elif cell_type == "tetrahedron":
+        edge_count = 6
+        face_count = 4
+        tdim = 3
+    elif cell_type == "hexahedron":
+        edge_count = 12
+        face_count = 6
+        tdim = 3
+
+    if tdim >= 2:
+        if tdim == 3:
+            face_start = 3 * face_count
+        else:
+            face_start = 0
+
+        dofstart = 0
+        for i in entity_dofs[0]:
+            dofstart += i
+
+        for e in range(edge_count):
+            edofs = entity_dofs[1][e]
+            if cell_info >> (face_start + e) & 1:
+                for b in range(block_size):
+                    s = (dofstart * block_size + b, (dofstart + edofs) * block_size, block_size)
+                    data[slice(*s)] = entity_transformations[0].dot(data[slice(*s)])
+            dofstart += edofs
+
+        if tdim == 3:
+            for f in range(face_count):
+                fdofs = entity_dofs[2][f]
+                if cell_info >> (3 * f) & 1:
+                    for b in range(block_size):
+                        s = (dofstart * block_size + b, (dofstart + fdofs) * block_size, block_size)
+                        data[slice(*s)] = entity_transformations[2].dot(data[slice(*s)])
+                for _ in range(cell_info >> (3 * f + 1) & 3):
+                    for b in range(block_size):
+                        s = (dofstart * block_size + b, (dofstart + fdofs) * block_size, block_size)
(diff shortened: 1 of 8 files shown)


python/basix/numba_helpers.py
```diff
@@ -0,0 +1,56 @@
+try:
+    import numba
+except ImportError:
+    raise RuntimeError("You must have numba installed to use the numba helper functions.")
+
+
+@numba.njit
+def apply_dof_transformation(entity_transformations, cell_type, entity_dofs,
```
garth-wells, line 8: Could you add a docstring? It's not so clear which arguments are in and which are out.

python/basix/numba_helpers.py
```diff
@@ -0,0 +1,56 @@
...
+                             data, block_size, cell_info):
+    if cell_type == "interval":
```
garth-wells, line 10: String comparison isn't so elegant - maybe return different functions depending on the cell type which can then be used without the internal check.

python/basix/numba_helpers.py
```diff
@@ -0,0 +1,56 @@
...
+            face_start = 3 * face_count
+        else:
+            face_start = 0
+
+        dofstart = 0
+        for i in entity_dofs[0]:
+            dofstart += i
+
+        for e in range(edge_count):
+            edofs = entity_dofs[1][e]
```
garth-wells, line 38: What type does `entity_dofs` normally have? (Will it be fast?)

garth-wells (commented):
Nice, just a few inline comments on docs and performance.

michalhabera:
But why is not the c++ `apply_dof_transformation` exposed and used in python?

mscroggs:
> But why is not the c++ `apply_dof_transformation` exposed and used in python?

I don't think you can call that from inside a numba compiled kernel?

michalhabera:
> > But why is not the c++ `apply_dof_transformation` exposed and used in python?
> 
> I don't think you can call that from inside a numba compiled kernel?

Right. Its just very dangerous to have the same algorithm implemented in two places, error prone. This would be possible if basix had C interface for methods which are meant to be called in the tight loops in custom assemblers.

jorgensd:
> > > But why is not the c++ `apply_dof_transformation` exposed and used in python?
> > 
> > 
> > I don't think you can call that from inside a numba compiled kernel?
> 
> Right. Its just very dangerous to have the same algorithm implemented in two places, error prone. This would be possible if basix had C interface for methods which are meant to be called in the tight loops in custom assemblers.

We had a C interface until recently. The problem is that it was not tested, that causes new issues when implementing new functionality.
The dof permutation code is fairly short, and should not be too hard to maintain. I can add a test to dolfin-x using this functionality to assemble mass matrices if we want to test its correctness.

michalhabera:
> > > > But why is not the c++ `apply_dof_transformation` exposed and used in python?
> > > 
> > > 
> > > I don't think you can call that from inside a numba compiled kernel?
> > 
> > 
> > Right. Its just very dangerous to have the same algorithm implemented in two places, error prone. This would be possible if basix had C interface for methods which are meant to be called in the tight loops in custom assemblers.
> 
> We had a C interface until recently. The problem is that it was not tested, that caused new issues when implementing new functionality in dolfin-x.
> The dof permutation code is fairly short (~50 lines of code), and should not be too hard to maintain. I can add a test to dolfin-x using this functionality to assemble mass matrices if we want to test its correctness.

It is tested against standard c++ code, so that is safe. For the actual permutation case this is not an issue, but it shows an issue in general. Consider as a next step I'd like to call basix basis evaluation inside my custom numba assembler - which is very much desired and one of the advantages for runtime tabulation library, right? Will you reimplement basis evaluation in numba?

If there is a way to call basix methods from numba directly this would be neat.

python/basix/numba_helpers.py
jorgensd: entity_dofs[i] is an array of ints. Input type into function has been changed in the doc strings

jorgensd (approved):

mscroggs:
I've made an issue for wider Numba support: #201

PR #199, opened 2021-04-20
Add CMake option to enable SIMD with xtensor

garth-wells:


IgorBaratta:
Should we add a message below `if (XTENSOR_USE_XSIMD AND xsimd_FOUND)`
Something like 
` message(" xsimd enabled, using version ${xsimd_VERSION}")`?

garth-wells:
Will reopen shortly.

PR #205, opened 2021-04-22
Version from cmake

chrisrichardson:
* Uses the version from the CMakeLists.txt to fill in `version.h` 
* Python package still needs to be updated separately, however - I can't think of any reasonably simple way to avoid it.
* Bumped version to 0.1.0 - this can be the "working version" until we actually tag it.


cpp/basix/version.h.cmake
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2020 Chris Richardson
```
garth-wells, line 1: I'm confused by the name of this file. Doesn't look like a CMake file. Should it be `version.h.in` (a template)?

A possibly simpler approach is to just put the versions numbers in `version.h` and let CMake parse the versions, see https://github.com/xtensor-stack/xtensor/blob/master/CMakeLists.txt#L15.

PR #216, opened 2021-04-30
Remove legacy setup.py things

jhale:
- Add Classifiers, makes for a nice pypi page.
- Require Python 3.7, no testing taking place on 3.6.
- Remove claims of Solaris and Windows support.
- Remove Windows specific lines.
- Add author and maintainer emails consistent with ffcx.


python/setup.py
```diff
@@ -43,38 +55,33 @@ def build_extension(self, ext):
...
+        if not os.path.exists(self.build_temp):
+            os.makedirs(self.build_temp)
+        subprocess.check_call(['cmake', ext.sourcedir] + cmake_args, cwd=self.build_temp, env=env)
+        subprocess.check_call(['cmake', '--build', '.'] + build_args, cwd=self.build_temp, env=env)
 
 
 setup(name='fenics-basix',
       version='0.1.0',
+      description='Basix Python interface',
+      url="https://github.com/fenics/basix",
```
chrisrichardson, line 76: I suppose it is the same, but we usually use FEniCS is the URL

chrisrichardson (approved):

PR #222, opened 2021-05-17
Implement geometry functions

### LINKED ISSUE #210: Provide functions that compute information about reference element
Currently various geometry data is computed by FFCx in FEniCS/ffcx#335. These are:

- [x] normals of facets of reference
- [x] orientations of normals (inward or outward)
- [x] Volume of reference
- [x] Volume of reference's facets
- [ ] Vectors of edges of cell
- [ ] Vectors of edges of cell's facets

In particular, it would be very useful for Basix to provide the user with the normals to facets (and orientations) and the volume of a reference, as these are a little involved to compute. The others are easy enough to calculate so can probably stay in FFCx.

mscroggs:
Fixes #210.

### DIFF
--- a/cpp/basix/cell.cpp
+++ b/cpp/basix/cell.cpp
@@ -5,6 +5,7 @@
 #include "cell.h"
 #include "quadrature.h"
 #include <map>
+#include <xtensor-blas/xlinalg.hpp>
 #include <xtensor/xview.hpp>
 
 using namespace basix;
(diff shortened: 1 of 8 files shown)


cpp/basix/cell.cpp
```diff
@@ -247,3 +248,182 @@ const std::string& cell::type_to_str(cell::type type)
...
+}
+//-----------------------------------------------------------------------------
+xt::xtensor<double, 2> cell::facet_normals(cell::type cell_type)
+{
+  const int tdim = cell::topological_dimension(cell_type);
+  xt::xtensor<double, 2> geometry = cell::geometry(cell_type);
+  std::vector<std::vector<int>> facets = cell::topology(cell_type)[tdim - 1];
+
+  xt::xtensor<double, 2> normals({facets.size(), (std::size_t)tdim});
+  for (std::size_t facet = 0; facet < facets.size(); ++facet)
```
garth-wells, line 295: It seems a bit strange to me having the if statements inside the loop. It would be more natural to use a switch, with loops for each switch.

PR #224, opened 2021-05-19
Use a switch instead of ifs inside a for

mscroggs:


garth-wells:
See https://github.com/FEniCS/basix/pull/225.

PR #238, opened 2021-06-21
Add to_transpose numba function

mscroggs:
This is needed in Numba kernels now that transformations are performed outside the FFCx kernels


python/basix/numba_helpers.py
```diff
@@ -206,3 +206,175 @@ def apply_dof_transformation_pyramid(
...
+        The DOF transformations for each entity.
+    entity_dofs : Dict(ndarray(int32))
+        The number of DOFs on each entity.
+    data : np.array
+        The data. This will be changed by this function.
+    cell_info : int
+        An integer representing the orientations of the subentities of the cell.
+    """
+    apply_dof_transformation_to_transpose(2, 3, 1, entity_transformations, entity_dofs,
+                                          data, cell_info, List.empty_list(types.string))
```
garth-wells, line 275: What's the last argument?

python/basix/numba_helpers.py
```diff
@@ -206,3 +206,175 @@ def apply_dof_transformation_pyramid(
...
+        The DOF transformations for each entity.
+    entity_dofs : Dict(ndarray(int32))
+        The number of DOFs on each entity.
+    data : np.array
+        The data. This will be changed by this function.
+    cell_info : int
+        An integer representing the orientations of the subentities of the cell.
+    """
+    apply_dof_transformation_to_transpose(3, 6, 4, entity_transformations, entity_dofs,
+                                          data, cell_info, List(["triangle"] * 4))
```
garth-wells, line 317: What is `List(["triangle"] * 4)`?

python/basix/numba_helpers.py
mscroggs: docstring added

PR #240, opened 2021-06-21
Add entity_closure_dofs

### LINKED ISSUE #235: Add entity closure dofs and entity dofs to Finite element.
This is currently constructed in Dolfinx, But should really be in Basix.

mscroggs:
- Added `entity_closure_dofs` function
- Renamed `entity_dofs` to `entity_dofs_count`
- Added new `entity_dofs` that returns a list of DOF numbers on each subentity

Fixes #235.

This PR requires corresponding changes in FFCx (FEniCS/ffcx#365) and DOLFINx (FEniCS/dolfinx#1582)


cpp/basix/cell.h
```diff
@@ -35,6 +35,12 @@ xt::xtensor<double, 2> geometry(cell::type celltype);
 /// @return List of topology (vertex indices) for each dimension (0..tdim)
 std::vector<std::vector<std::vector<int>>> topology(cell::type celltype);
 
+/// Get the numbers of entitys connected to each subentity of the cell
+/// @param celltype Cell Type
+/// @return List of topology (vertex indices) for each dimension (0..tdim)
+std::vector<std::vector<std::vector<std::vector<int>>>>
```
garth-wells, line 41: Could you expand on the doc for the return object? It has four indices - what is each one?

garth-wells (changes requested):
Just a query on a return type.

cpp/basix/cell.h
mscroggs: I've added to to doc

garth-wells (approved):

Issue #251, opened 2021-08-02
Implement new lattice points on simplex cells


mscroggs:
On simplices, GLL points lead to elements with a lower Lebesgue constant than equispaced points, but the Lebesgue constant still grow rapidly the polynomial order is increased:

![Figure_1](https://user-images.githubusercontent.com/9850599/127827992-fe3a00c5-d1a4-4054-ab39-a752c3bf11ce.png)

The Lebesgue constant for elements generated with Fekete points will be better behaved, although it seems that an optimisation problem needs to be solved to find these, and they are often not symmetric. A set of (preferably symmetric) points leading to elements with a good enough Lebesgue constant are desired.

mscroggs:
> Toby Isaac has a simple symmetric construction with good properties for simplex nodes of all dimensions in this preprint https://arxiv.org/abs/2002.09421

Thanks, this saves me a lot of searching and reading

Issue #72, opened 2020-11-23
Support "variants" of elements


chrisrichardson:
Do we have "variants" of elements, e.g. Lagrange at GLL points? Discontinuous Lagrange? Point eval vs. integral moment?

Or should each variant just have a new name?

PR #252, opened 2021-08-02
Allow use of GLL points for Lagrange

### LINKED ISSUE #16: Use GLL points for tensor product Lagrange elements
Default to GLL points rather than equispaced points.

### LINKED ISSUE #65: Use shorter family names
Use names which are shorter, e.g. from FEEC.

### LINKED ISSUE #232: Support Lagrange elements on interval with GLL points
Currently, it's possible to get GLL quadrature points, but not the degrees of freedom on GLL points.
The goal is the implementation of high-order tensor-product elements on hexes.

mscroggs:
Adds an extra argument to `create_lagrange` that gives that lattice type that should be used for the layout of the DOF points. ~By default GLL points will be used.~ See #251 for plots showing the improved Lebesgue constants when using GLL.

Overloads `create_element`, adding a new version that takes a lattice type input and passes this on to `create_lagrange` or throws an error for non-Lagrange elements.

Fixes #16. Fixes #232. Progress towards #72.

This is coupled with FEniCS/ffcx#376 and FEniCS/dolfinx#1651.

Update:

Removed the versions of `create_element` in C++ that accept strings.
Added a create_element function to the Python interface that accepts alterntive names for elements, allowing the name mapping in FFCx to be simplified.
Shotened the names used for elements: fixes #65


cpp/basix/finite-element.cpp
```diff
@@ -74,6 +74,14 @@ constexpr int num_transformations(cell::type cell_type)
 }
 } // namespace
 //-----------------------------------------------------------------------------
+basix::FiniteElement basix::create_element(std::string family, std::string cell,
```
garth-wells, line 77: Could we remove this function and just have the version that takes an enum?

cpp/basix/finite-element.h
```diff
@@ -716,12 +717,21 @@ class FiniteElement
       _etrans_invT;
 };
 
+/// Create an element by name
+FiniteElement create_element(std::string family, std::string cell, int degree,
+                             lattice::type lattice_type);
+
 /// Create an element by name
 FiniteElement create_element(std::string family, std::string cell, int degree);
```
garth-wells, line 725: How about removing this version?

cpp/basix/finite-element.h
```diff
@@ -716,12 +717,21 @@ class FiniteElement
...
 
 /// Create an element by name
-FiniteElement create_element(element::family family, cell::type cell,
-                             int degree);
+FiniteElement
```
garth-wells, line 728: Add docstring, including description for each argument.

cpp/basix/lagrange.h
```diff
@@ -6,25 +6,29 @@
 
 #include "cell.h"
 #include "finite-element.h"
+#include "lattice.h"
 
 namespace basix
 {
 /// Create a Lagrange element on cell with given degree
-/// @param celltype interval, triangle, quadrilateral, tetrahedral, or
+/// @param[in] celltype interval, triangle, quadrilateral, tetrahedral, or
```
garth-wells, line 14: Add brief description of what `celltype` is - details of `lattice::type` will be documented where `lattice::type` is declared.

cpp/basix/lagrange.h
```diff
@@ -6,25 +6,29 @@
...
 /// hexahedral celltype
 /// @param[in] degree
+/// @param[in] lattice_type
 /// @return A FiniteElement
-FiniteElement create_lagrange(cell::type celltype, int degree);
+FiniteElement create_lagrange(cell::type celltype, int degree,
+                              lattice::type lattice_type
+                              = lattice::type::gll_warped);
```
garth-wells, line 21: Maybe remove the default to force caller to think about this?

garth-wells (changes requested):

cpp/basix/finite-element.cpp
mscroggs: As different elements might need to be passed different enums (eg Nedelec might want a polynomial space type to take moments against (legendre, chebyshev, etc), rather than a point arrangement) so I thought it could be neater to overload `create_element` and throw errors when the wrong additional arguments are passed to an element.

We may be able to get rid of some though, as having string and element::family/cell::type versions of each feels like it might be too much

garth-wells: Let's boil it down to the minimum interface - can always add interfaces later if required.

mscroggs: We have this issue already though, as Lagrange spaces need a lattice::type, and other elements don't. It feels bad for `create_element` to always take this input, but for it to have no effect for other spaces

garth-wells: Happy to have extra interfaces that for extra arguments - let's just remove redundant interfaces, e.g. remove strings in favour of enums.

cpp/basix/lattice.h
```diff
@@ -16,6 +16,12 @@ enum class type
   gll_warped
 };
 
+/// Convert string to a lattice type
+lattice::type str_to_type(std::string name);
+
+// Convert family to string
+const std::string& type_to_str(lattice::type type);
```
garth-wells, line 23:
```suggestion
std::string type_to_str(lattice::type type);
```

garth-wells (approved):

PR #260, opened 2021-08-16
Allow discontinuous elements

### LINKED ISSUE #203: `cannot create std::vector larger than max_size()` when trying to make Lagrange 0
I get this error when creating an order 0 Lagrange space on a triangle, tet, quad or hex
```
import basix

element = basix.create_element("Lagrange", "triangle", 0)
```

Should be replaced with a better error message

### LINKED ISSUE #245: Remove Discontinuous Lagrange element
Having different ordering of DOFs for continuous and discontinuous seems like an overcomplication

### LINKED ISSUE #246: Allow discontinuous spaces
For any space, the discontinuous version should be obtainable. This should have the same DOFs, but will all associated with the interior rather than vertices/edges/faces.

### LINKED ISSUE #261: Discontinuous Raviart-Thomas

mscroggs:
Added a `discontinuous` input to `create_element` that makes a version of the element with all the DOFs associated with the interior of the cell.

Added `e-` prefix to element files to make it easier to tell which files are which.

This PR is coupled with FEniCS/dolfinx#1670 and FEniCS/ffcx#382.

Fixes #245. Fixes #246. Fixes #203. Fixes #261.


cpp/basix/finite-element.h
```diff
@@ -188,6 +188,16 @@ xt::xtensor<double, 3> compute_expansion_coefficients(
     const std::vector<std::vector<xt::xtensor<double, 2>>>& x, int degree,
     double kappa_tol = 0.0);
 
+/// TODO: doc
```
mscroggs, line 191: Need to replace this TODO with documentation

cpp/basix/e-crouzeix-raviart.cpp
```diff
@@ -63,6 +64,13 @@ FiniteElement basix::create_cr(cell::type celltype, int degree)
   }
 
   M[tdim - 1].resize(facet_topology.size(), xt::ones<double>({1, 1, 1}));
+
+  if (discontinuous)
```
garth-wells, line 68: Could this be moved higher up?

There are some more cases below where I think the exception can be raised earlier - it would be good to place it as early as possible in functions.

cpp/basix/e-lagrange.cpp
```diff
@@ -34,6 +35,9 @@ FiniteElement basix::create_lagrange(cell::type celltype, int degree,
   // Create points at nodes, ordered by topology (vertices first)
   if (degree == 0)
   {
+    if (!discontinuous)
```
garth-wells, line 38: Add braces.

cpp/basix/e-crouzeix-raviart.cpp
mscroggs: This one can't as it needs x, M, and entity_transformations to be made before it so they can then be updated to make the discontinuous element. I've moved up the exceptions that could be moved though

cpp/basix/finite-element.cpp
```diff
@@ -256,21 +256,81 @@ xt::xtensor<double, 3> basix::compute_expansion_coefficients(
...
+          .assign(M[i][j]);
+      x_n += x[i][j].shape(0);
+      M_n += M[i][j].shape(0);
+    }
+  }
+
+  x_out[tdim].push_back(new_x);
+  M_out[tdim].push_back(new_M);
+
+  for (auto i = entity_transformations.begin();
```
garth-wells, line 308: `auto it =` would more conventional. `i` is so commonly used as an integer index a reader could easily assume that it is.

cpp/basix/finite-element.cpp
```diff
@@ -256,21 +256,81 @@ xt::xtensor<double, 3> basix::compute_expansion_coefficients(
...
+       i != entity_transformations.end(); ++i)
+  {
+    entity_transformations_out[i->first]
+        = xt::xtensor<double, 3>({i->second.shape(0), 0, 0});
+  }
+
+  return std::make_tuple(x_out, M_out, entity_transformations_out);
```
garth-wells, line 315:
```suggestion
  return {x_out, M_out, entity_transformations_out};
```

cpp/basix/finite-element.h
```diff
@@ -289,6 +313,11 @@ class FiniteElement
   /// @return The mapping
   maps::type mapping_type() const;
 
+  /// Indicates whether this element is the discontinuous variant
+  /// @return A bool indicating whether or not this is the discontinuous version
```
garth-wells, line 317:
```suggestion
  /// @return True is this element is a discontinuous version
```

cpp/basix/finite-element.h
```diff
@@ -725,24 +754,30 @@ class FiniteElement
            std::vector<std::tuple<std::vector<std::size_t>, std::vector<double>,
                                   xt::xtensor<double, 2>>>>
       _etrans_invT;
+
+  /// Indicates whether or not this is the discontinuous version of the element
```
garth-wells, line 758:
```suggestion
  // Indicates whether or not this is the discontinuous version of the element
```
Double forward slash only for private variables.

cpp/basix/finite-element.h
```diff
@@ -725,24 +754,30 @@ class FiniteElement
...
 /// Create an element using a given lattice type
 /// @param[in] family The element family
 /// @param[in] cell The reference cell type that the element is defined on
 /// @param[in] degree The degree of the element
 /// @param[in] lattice_type The lattice type that should be used to arrange DOF
-/// points of the element
-FiniteElement
-create_element(element::family family, cell::type cell, int degree,
-               lattice::type lattice_type);
+/// @param[in] discontinuous Indicates whether the element is discontinuous
```
garth-wells, line 767: Could we add a brief comment on what different between continuous and discontinuous is in practice? E.g. comment in the dof associations.

cpp/basix/finite-element.h
```diff
@@ -725,24 +754,30 @@ class FiniteElement
...
+/// between cells points of the element
+FiniteElement create_element(element::family family, cell::type cell,
+                             int degree, lattice::type lattice_type,
+                             bool discontinuous = false);
```
garth-wells, line 771: Do we want the default arg at the end, or should be required it to be added?

cpp/basix/finite-element.h
```diff
@@ -725,24 +754,30 @@ class FiniteElement
...
 /// Create an element
 /// @param[in] family The element family
 /// @param[in] cell The reference cell type that the element is defined on
 /// @param[in] degree The degree of the element
-FiniteElement
-create_element(element::family family, cell::type cell, int degree);
+/// @param[in] discontinuous Indicates whether the element is discontinuous
+/// between cells
+FiniteElement create_element(element::family family, cell::type cell,
+                             int degree, bool discontinuous = false);
```
garth-wells, line 780: In my experience default arguments like this can be quit dangerous because `int` can be silently cast to `bool`. User could make a mistake to the arguments and it won't het picked up by the compiler.

garth-wells (changes requested):
Nice.

cpp/basix/finite-element.h
mscroggs: I've removed the default value

garth-wells (approved):

PR #274, opened 2021-09-15
Drop xtensor-linalg dependency

IgorBaratta:
WIP:
 - solve
 - cond

### DIFF
--- a/cpp/basix/math.h
+++ b/cpp/basix/math.h
@@ -44,4 +44,23 @@ xt::xtensor_fixed<typename U::value_type, xt::xshape<3>> cross(const U& u,
           u[0] * v[1] - u[1] * v[0]};
 }
 
+/// Compute C = A * B
+/// @param[in] A Input matrix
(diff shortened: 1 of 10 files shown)


cpp/basix/math.h
```diff
@@ -0,0 +1,118 @@
...
+/// Compute the eigenvalues and eigenvectors of a square Hermitian matrix A
+/// @param[in] A Input matrix
+/// @param[out] C Filled to be C = A * B
+template <typename U>
+auto eigh(const U& A)
+{
+  // Copy to column major matrix
+  xt::xtensor<double, 2, xt::layout_type::column_major> M(A.shape());
+  M.assign(A);
+  int N = A.shape(0);
```
mscroggs, line 81: You could change this to something like `std::array<int> N = {A.shape(0)};` then remove the {}s from the next line. I think that'll fix the error its currently hitting when compiling.

garth-wells:
Taken forward in #291.

PR #279, opened 2021-09-20
Add Chebyshev point lattice

mscroggs:
Added Chebyshev points as a lattice to be used for creating Lagrange elements. This new variant can only be used for discontinuous Lagrange elements, as the endpoints are not included in the Chebyshev points.

For simplices, the Chebyshev points are created by defining the Chebyshev points on an interval, adding the end points then using warping or Isaac's method to create points in the cell. (This can probably be improved on)

The Lebesgue constants for these points (compared to equispaced and GLL points) are shown here:

Figure_4

Other changed in this PR that may need discussing before merging:

- Replaced `lattice::type` with `lattice::type` and `lattice::simplex_method`, as each lattice type can have a warped or isaac variant and it made sense to me to split the point spacing on an interval from the method used to turn this into points on non-TP cells.
- Added `element::lagrange_variant` and made Lagrange elements take this as in input instead of a lattice type. This is because (eg) Chebyshev points are ordered differently so taking the Chebyshev variant does more than just change the lattice type. It also would allow us to support other variants of Lagrange (eg. variant where basis functions are orthonormal polynomials, variant with integral moments) where the change is not just due to a different Lattice.

### DIFF
--- a/cpp/basix/e-brezzi-douglas-marini.cpp
+++ b/cpp/basix/e-brezzi-douglas-marini.cpp
@@ -39,8 +39,8 @@ FiniteElement basix::create_bdm(cell::type celltype, int degree,
   std::array<std::vector<xt::xtensor<double, 2>>, 4> x;
 
   // Add integral moments on facets
-  const FiniteElement facet_moment_space
-      = create_lagrange(facettype, degree, lattice::type::equispaced, true);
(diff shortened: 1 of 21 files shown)


garth-wells (changes requested):
Looks good. Just some small suggestions.

PR #286, opened 2021-09-24
Builds and tests manylinux wheels for various x86-64 Pythons

jhale:
Still need to add GitHub Actions integration.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,3 +2,28 @@
 requires = ["setuptools>=42", "wheel", "pybind11>=2.6.4", "cmake>=3.16", "scikit-build>=0.12"]
 
 build-backend = "setuptools.build_meta"
+
+[tool.cibuildwheel]
+build = ["cp*-manylinux_x86_64"]
+skip = ""
+test-skip = ""
+
+archs = ["auto"]
+build-frontend = "pip"
+dependency-versions = "pinned"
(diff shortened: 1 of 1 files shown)


mscroggs:
Do you want to add the actions integration in the PR or should we merge this and do that later?

garth-wells:
We should wait on #274 before creating wheel - it remove many files that would otherwise be 'baked' in.

garth-wells:
Converted to draft so we don't merge before #274 .

jhale:
https://github.com/FEniCS/basix/pull/286/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R9

This test won't run because the temporary build path is written into the various cmake config files. The wheels are broken with respect to building a downstream C++ library on top.

jhale:
Close on Linux now. Still need to test the MacOS wheels.

On Linux, there is an outstanding issue with the linker not being able to find the wheel-packaged libgfortran. The library is there in `basix.libs/`.

https://github.com/FEniCS/basix/pull/286/checks?check_run_id=3775702326#step:3:510

jhale:
Even though the wheel contains the folder `fenics_basix.libs`, these are not installed into `basix/` when running `pip install *.whl`. So no surprise that there is an issue.

PR #291, opened 2021-09-30
Remove xtensor-blas dependency

garth-wells:


test/test_dof_transformations.py
```diff
@@ -9,7 +9,7 @@
 from .utils import parametrize_over_elements
 
 
-@parametrize_over_elements(3)
+@parametrize_over_elements(1)
```
mscroggs, line 12: It would be good to test more than order 1 elements here, as lots of elements have trivial transformations for lowest order but not for higher order.

mscroggs (approved):

IgorBaratta (approved):
Only one comment/question:
Do we need to worry about integer types?

PR #292, opened 2021-10-01
Add interpolation between elements

mscroggs:
Added `compute_interpolation_between_elements` that computes a matrix that represents the interpolation between two elements on the reference cell


jhale:
The interpolation of vector Lagrange into H(curl) or H(div) is a practically relevant test case.

mscroggs:
Yes, once interpolation in DOLFINx uses this matrix, we can very easily map (eg) Nedelec to Vector DG for visualisation. I'm just updating the docs to document the fact that the function does this

IgorBaratta (approved):

PR #316, opened 2021-10-29
Allow DG0 on a point

### LINKED ISSUE #209: Support DG 0 elements for points
To be able to support point cloud data, similar to what was presented by: https://fenics2021.com/talks/nixon-hill.html
we need to be able to make DG-0 elements using basix (which can then be used to make a point-cloud mesh in dolfinx).

mscroggs:
Fixes #209.


cpp/basix/cell.cpp
```diff
@@ -16,6 +16,8 @@ xt::xtensor<double, 2> cell::geometry(cell::type celltype)
 {
   switch (celltype)
   {
+  case cell::type::point:
+    return xt::xtensor<double, 2>({{}});
```
garth-wells, line 20: Should this be
```
return xt::xtensor<double, 2>({{0.0}});
```
?

cpp/basix/cell.cpp
```diff
@@ -59,6 +61,11 @@ std::vector<std::vector<std::vector<int>>> cell::topology(cell::type celltype)
 
   switch (celltype)
   {
+  case cell::type::point:
+    t.resize(1);
+    // Vertices
+    t[0] = {{}};
```
garth-wells, line 67: Should this be `t[0] = {{0}};`?

garth-wells (commented):
Few queries inline on point handling.

cpp/basix/cell.cpp
mscroggs: Point has 0 topological dimension so I thought should have coordinates of length 0. But I'm not certain this is sensible

mscroggs: Tests pass either way, so I'm adding the 0.0 to make it length 1. Can always revise it once we have an example in DOLFINx that actually uses this functionality

PR #322, opened 2021-11-02
Add simple cpp demo

IgorBaratta:
Shows how to link basix (cmake),  and useful for testing in hpc enviroment.


mscroggs (approved):
looks good

.github/workflows/dolfin-tests.yml
```diff
@@ -39,6 +39,13 @@ jobs:
           python3 -m pip install git+https://github.com/FEniCS/ufl.git
           python3 -m pip install git+https://github.com/FEniCS/ffcx.git
 
+      - name: Run cpp demos
```
mscroggs, line 42: Is this intentionally in the DOLFINx tests workflow? Would it make more sense in pythonapp.yml?

mscroggs:
I get `/usr/local/lib/libbasix.so: undefined reference to `dladdr'` when I try to make the demo locally

.github/workflows/dolfin-tests.yml
IgorBaratta: pythonapp.yml installs basix with pip, so it's not installed in the default foder.

IgorBaratta:
> I get `/usr/local/lib/libbasix.so: undefined reference to `dladdr'` when I try to make the demo locally

I've updated the CMakeLists.txt, could you check if it fixes the issue?

mscroggs (approved):

PR #323, opened 2021-11-02
Fix for 'maybe unused' variable in release mode

jorgensd:
`tdim` is only used in an assertion, and will there be unused if running in `Release` mode.

### DIFF
--- a/cpp/basix/moments.cpp
+++ b/cpp/basix/moments.cpp
@@ -395,9 +395,7 @@ moments::make_dot_integral_moments(const FiniteElement& V, cell::type celltype,
   auto wts = xt::adapt(_wts);
 
   // If this is always true, value_size input can be removed
-  [[maybe_unused]] const std::size_t tdim
-      = cell::topological_dimension(celltype);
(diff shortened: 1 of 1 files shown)


mscroggs (approved):

cpp/basix/moments.cpp
```diff
@@ -389,13 +389,14 @@ moments::make_dot_integral_moments(const FiniteElement& V, cell::type celltype,
...
-  const std::size_t tdim = cell::topological_dimension(celltype);
 
   auto [pts, _wts] = quadrature::make_quadrature(quadrature::type::Default,
                                                  sub_celltype, q_deg);
   auto wts = xt::adapt(_wts);
 
   // If this is always true, value_size input can be removed
+  [[maybe_unused]] const std::size_t tdim
+      = cell::topological_dimension(celltype);
   assert(tdim == value_size);
```
garth-wells, line 400: Better to use 
```cpp
assert(cell::topological_dimension(celltype) == value_size);
```

garth-wells (approved):

PR #333, opened 2021-11-16
Remove `compute_expansion_coefficients` and store dual matrix

mscroggs:
Removed the `compute_expansion_coefficients` function, and move the computation it did inside the `FiniteElement` constructor.

Removed `_matM_new` from `FiniteElement` as is it unused.

Added `dual_matrix()` and `coefficient_matrix()` methods so these matrices can be accessed from Python

### DIFF
--- a/cpp/basix/cell.cpp
+++ b/cpp/basix/cell.cpp
@@ -15,7 +15,7 @@ xt::xtensor<double, 2> cell::geometry(cell::type celltype)
   switch (celltype)
   {
   case cell::type::point:
-    return xt::xtensor<double, 2>({{0.0}});
+    return xt::xtensor<double, 2>({{}});
   case cell::type::interval:
     return xt::xtensor<double, 2>({{0.0}, {1.0}});
   case cell::type::triangle:
--- a/cpp/basix/finite-element.h
+++ b/cpp/basix/finite-element.h
@@ -763,6 +763,8 @@ public:
   /// @return The dual matrix
   xt::xtensor<double, 2> dual_matrix() const;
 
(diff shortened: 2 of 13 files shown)


cpp/basix/finite-element.h
```diff
@@ -773,6 +756,15 @@ class FiniteElement
...
                    const xtl::span<const T>& data, const int block_size) const;
 
+  /// Get the dual matrix.
+  ///
+  /// This is the matrix @f$BD^{T}@f$, as described in the documentation of the
+  /// `FiniteElement()` constructor.
+  /// @return The dual matrix
+  xt::xtensor<double, 2> dual_matrix() const;
+
+  xt::xtensor<double, 2> coefficient_matrix() const;
```
garth-wells, line 766: Doc string is missing.

cpp/basix/cell.cpp
```diff
@@ -15,7 +15,7 @@ xt::xtensor<double, 2> cell::geometry(cell::type celltype)
   switch (celltype)
   {
   case cell::type::point:
-    return xt::xtensor<double, 2>({{0.0}});
+    return xt::xtensor<double, 2>({{}});
```
garth-wells, line 18: Is this change intended?

mscroggs: Maybe.

I think the shape of `cell::geometry` should be npoints by tdim, so in this case 1 by 0.

On the other hand, this isn't the change that fixed the space on points, so we could alternatively have the shape as npoints by gdim and declare that the gdim for a point is 1.

garth-wells (approved):

Issue #254, opened 2021-08-06
Change quadrature rules


jorgensd:
In the package quadpy, there is a large collection of quadrature rules.
https://github.com/nschloe/quadpy
i would suggest replacing some of the existing rules with rules from here. The rules are stored as json files.

mscroggs:
Progress towards this was made in #258

garth-wells:
@jorgensd Which rules in particular?

jorgensd:
Whenever we fall back to an `m`th order Gauss-Jacobi rule (https://github.com/FEniCS/basix/blob/main/cpp/basix/quadrature.cpp#L1736-L1759), we use:
- Interval: `(m+2) / 2`
- Quadrilateral `((m+2) / 2)**2`
- Hexahedron `((m+2) / 2)**3`
- Triangle: `((m+2) / 2)**2`
- Tetrahedron: `((m+2) / 2)**3`
Meaning that for a 9th order quadrature rule on a tetrahedron, we have 125 quadrature points while Xiao-Gimbutas only uses 57  points (as far as I can tell). https://github.com/nschloe/quadpy/tree/main/src/quadpy/t3/_xiao_gimbutas
The goal would be to get all the default results to a relatively high order to use other points than Gauss-Jacobi

mscroggs:
I think we have good schemes on intervals, quads, hexes, and triangles. Once we add some rules for tets up to a reasonable order, I think this is done

PR #338, opened 2021-11-18
Reduce quadrature orders used in element creation

mscroggs:
Reduce the quadrature orders used when defining integral moment DOFs, leading elements having lower numbers of interpolation points.

The number of interpolation points (particularly on tetrahedra) could be further reduced by implementing more quadrature schemes with lower numbers of points (see #254)

Elements on simplices
Figure_1

Elements on tensor products
Figure_2

### DIFF
--- a/.github/workflows/dolfin-tests.yml
+++ b/.github/workflows/dolfin-tests.yml
@@ -48,7 +48,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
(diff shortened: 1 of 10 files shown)


garth-wells (approved):

mscroggs:
The extra tests turned on in https://github.com/FEniCS/dolfinx/pull/1818 still pass following these changes, so I'm happy they don't break any spaces

PR #346, opened 2021-11-26
Add information about highest degree of polynomials in a finite element

### LINKED ISSUE #345: Add information about highest order polynomial in space
Can be used to find appropriate interpolation spaces.

mscroggs:
Store two number in FiniteElement:

- `highest_polynomial_degree` gives the degree of the Lagrange space that the element is a subspace of.
- `highest_complete_polynomial_degree` gives the degree of the Lagrange space that is a subspace of the element.

`highest_polynomial_degree` will be particularly useful as it tells us which degree Lagrange space to interpolate into for visualising a function

Fixes #345.

### DIFF
--- a/cpp/basix/e-brezzi-douglas-marini.cpp
+++ b/cpp/basix/e-brezzi-douglas-marini.cpp
@@ -75,6 +75,7 @@ FiniteElement basix::element::create_bdm(cell::type celltype, int degree,
 
   return FiniteElement(element::family::BDM, celltype, degree, {tdim},
                        xt::eye<double>(ndofs), entity_transformations, x, M,
-                       maps::type::contravariantPiola, discontinuous);
+                       maps::type::contravariantPiola, discontinuous, degree,
+                       degree);
 }
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


mscroggs:
The names given to these two values may need more thought

garth-wells (commented):
Function names are a bit of a mouthful could we shorten? Would it work to have a function 
```cpp
std::array<int, 2> degree();
```
that returns (lower bound, upper bound) on the degree?

mscroggs:
> Would it work to have a function
> 
> ```c++
> std::array<int, 2> degree();
> ```
> 
> that returns (lower bound, upper bound) on the degree?

Good idea

garth-wells:
> Good idea, except we already have `degree()` that returns the degree passed in when creating the space

Which degree do we pass in - the 'lower' or 'upper' bound?

mscroggs:
> Which degree do we pass in - the 'lower' or 'upper' bound?

Depends on the element. Ususally we pass in the lower bound, but for some elements (eg serendipity) we pass in something else.

I'm calling this `degree_bounds` for now

garth-wells (approved):

PR #354, opened 2021-12-01
Enable spack installation on centos 8

IgorBaratta:
This is a known issue, but I don't know if we have a better solution at the moment.
This doesn't affect runtime performance, and the size of the pybind wrapper changes only slightly.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -46,7 +46,7 @@ find_package(pybind11 REQUIRED CONFIG HINTS ${PYBIND11_DIR} ${PYBIND11_ROOT}
   $ENV{PYBIND11_DIR} $ENV{PYBIND11_ROOT})
 
 # Create the binding library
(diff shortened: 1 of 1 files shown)


chrisrichardson:
I am not convinced this is a general problem on Centos8.

jhale:
Can you give some links to these known issues? Would be good to find out a bit more about the problem before turning this off.

IgorBaratta:
> Can you give some links to these known issues? Would be good to find out a bit more about the problem before turning this off.

I was not able to reproduce the issue on an isolated system.
With LTO basix (python) crashes on our local cluster with Centos 8.

As an work around I'm using ```-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF```.
I think this should be an option on spack.

jhale:
I know that this can be caused using mismatched compilers e.g. anaconda compilers to build system python libraries and executable, and system compiler to build the pybind11 library.

PR #357, opened 2021-12-07
Remove xtensor, make download via git explicit option

jhale:


drew-parsons:
One question: the patch uses `if(NOT DOWNLOAD_XTENSOR_LIBS)... find_package()`.  Do you mean the intention is to always download (if DOWNLOAD_XTENSOR_LIBS is set) even if the system already does  have the required xtensor libs?

jhale:
Indeed, the intention is that if the user explicitly asks for downloaded libraries, they want them used in the build.

Otherwise the canonical CMake behaviour looking for the libraries and exiting if they are not found is the default.

jhale:
For the scikit-build build in the root `pip install .` I've set the xtensor libraries to always be downloaded. I think this aimed at users who just want it to work. The split C++/Python install you have full control.

mscroggs (approved):

garth-wells (approved):

PR #393, opened 2022-03-03
Remove memory assignment in polyset

### LINKED ISSUE #313: Support tabulate with no internal heap (dynamic) memory allocation
To support fast basis function evaluation, Basix should support tabulation without internal heap (dynamic) memory allocation. The code that needs revision to support this is `polyset`.

We should pass 'scratch' memory, and provide a function that reports how much memory is required.

mscroggs:
Fixes #313

### DIFF
--- a/cpp/basix/polyset.cpp
+++ b/cpp/basix/polyset.cpp
@@ -645,26 +645,36 @@ void tabulate_polyset_quad_derivs(xt::xtensor<double, 3>& P, std::size_t n,
   std::fill(P.begin(), P.end(), 0.0);
   xt::view(P, idx(0, 0), xt::all(), quad_idx(0, 0)) = 1.0;
 
-  for (std::size_t ky = 0; ky <= nderiv; ++ky)
+  if (n == 0)
+    return;
+
+  { // scope
+    auto result = xt::view(P, idx(0, 0), xt::all(), xt::all());
+    xt::col(result, quad_idx(0, 1))
+        = (x1 * 2.0 - 1.0) * xt::col(result, quad_idx(0, 0));
+    for (std::size_t py = 2; py <= n; ++py)
+    {
+      const double a = 1.0 - 1.0 / static_cast<double>(py);
+      xt::col(result, quad_idx(0, py))
+          = (x1 * 2.0 - 1.0) * xt::col(result, quad_idx(0, py - 1)) * (a + 1.0)
+            - xt::col(result, quad_idx(0, py - 2)) * a;
+    }
+  }
+  for (std::size_t ky = 1; ky <= nderiv; ++ky)
   {
     // Get reference to this derivative
     auto result = xt::view(P, idx(0, ky), xt::all(), xt::all());
     auto result0 = xt::view(P, idx(0, ky - 1), xt::all(), xt::all());
-    for (std::size_t py = 1; py <= n; ++py)
+    xt::col(result, quad_idx(0, 1))
+        = (x1 * 2.0 - 1.0) * xt::col(result, quad_idx(0, 0))
+          + 2 * ky * xt::col(result0, quad_idx(0, 0));
+    for (std::size_t py = 2; py <= n; ++py)
     {
       const double a = 1.0 - 1.0 / static_cast<double>(py);
       xt::col(result, quad_idx(0, py))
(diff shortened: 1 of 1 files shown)


cpp/basix/polyset.cpp
```diff
@@ -612,62 +601,112 @@ void tabulate_polyset_pyramid_derivs(xt::xtensor<double, 3>& P, int n,
...
   assert(x.shape(1) == 2);
   const std::size_t m = (n + 1) * (n + 1);
   const std::size_t md = (nderiv + 1) * (nderiv + 2) / 2;
 
+  // Indexing for quadrilateral basis functions
+  auto quad_idx = [n](std::size_t px, std::size_t py) -> std::size_t
+  { return (n + 1) * px + py; };
+
   // Compute 1D basis
   const xt::xtensor<double, 1> x0 = xt::col(x, 0);
```
garth-wells, line 634: Could this using `auto` to avoid the creation of a ` xt::xtensor<double, 1>`?

cpp/basix/polyset.cpp
```diff
@@ -680,51 +719,134 @@ void tabulate_polyset_hex_derivs(xt::xtensor<double, 3>& P, std::size_t n,
   const std::size_t m = (n + 1) * (n + 1) * (n + 1);
   const std::size_t md = (nderiv + 1) * (nderiv + 2) * (nderiv + 3) / 6;
 
+  // Indexing for hexahedral basis functions
+  auto hex_idx
+      = [n](std::size_t px, std::size_t py, std::size_t pz) -> std::size_t
+  { return (n + 1) * (n + 1) * px + (n + 1) * py + pz; };
+
   // Compute 1D basis
   const xt::xtensor<double, 1> x0 = xt::col(x, 0);
```
garth-wells, line 728: Can we avoid dynamic allocation here?

cpp/basix/polyset.cpp
```diff
@@ -680,51 +719,134 @@ void tabulate_polyset_hex_derivs(xt::xtensor<double, 3>& P, std::size_t n,
...
-  xt::xtensor<double, 3> pz({static_cast<std::size_t>(nderiv + 1), x2.shape(0),
-                             static_cast<std::size_t>(n + 1)});
-  tabulate_polyset_line_derivs(pz, n, nderiv, x2);
-
-  // Compute basis
+
+  assert(x0.shape(0) > 0);
+  assert(x1.shape(0) > 0);
+  assert(x2.shape(0) > 0);
+  const auto X0 = x0 * 2.0 - 1.0;
```
garth-wells, line 735: These could be doing allocations. Not sure how lazy the evaluations are, or how to check.

cpp/basix/polyset.cpp
jorgensd: I think so.

jorgensd:
Longterm, we should aim to transpose the polyset basis [in/out] data, as we currently have a strided access of the data, such as:
```c++
xt::view(P, xt::all(), xt::all(), hex_idx(0, py, pz)) = 0.0;
```
```c++
auto result = xt::view(P, idx(0, 0, kz), xt::all(), xt::all());
if (kz == 0)
     xt::col(result, hex_idx(0, 0, 0)) = 1.0;
xt::col(result, hex_idx(0, 0, pz))
          = (x2 * 2.0 - 1.0) * xt::col(result, hex_idx(0, 0, pz - 1))
            * (a + 1.0);
 ```
 This will be a major change, as it would affect all element creation.

mscroggs:
> Longterm, we should aim to transpose the polyset basis [in/out] data, as we currently have a strided access of the data, such as:
> 
> ```c++
> auto result = xt::view(P, idx(0, 0, kz), xt::all(), xt::all());
> if (kz == 0)
>      xt::col(result, hex_idx(0, 0, 0)) = 1.0;
> xt::col(result, hex_idx(0, 0, pz))
>           = (x2 * 2.0 - 1.0) * xt::col(result, hex_idx(0, 0, pz - 1))
>             * (a + 1.0);
> ```
> 
> This will be a major change, as it would affect all element creation.

I've opened #395 for this

cpp/basix/polyset.cpp
mscroggs: I've removed these

cpp/basix/polyset.cpp
mscroggs: We now use auto instead

garth-wells (approved):

PR #405, opened 2022-04-01
Wrap elements to be used directly in UFL

### LINKED ISSUE #63: Custom element in dolfin
In #62, we make finite element class store a family name string. This will be used by dolfin to recontruct the element for tabulation.

For custom element (eg the elements in `test_create.py`), this will not work as the element cannot be constructed.

mscroggs:
This allows Basix elements, including custom elements, to be used with FFCx and DOLFINx.

This is coupled with https://github.com/FEniCS/ffcx/pull/470 and https://github.com/FEniCS/dolfinx/pull/2113.

Closes #63.


cpp/basix/finite-element.cpp
```diff
@@ -216,6 +219,9 @@ basix::FiniteElement basix::create_element(element::family family,
                                            element::dpc_variant dvariant,
                                            bool discontinuous)
 {
+  if (family == element::family::custom)
+    throw std::runtime_error("Cannot create a custom element directly. Try "
```
garth-wells, line 223: Curly braces around the multi-line statement would help readability here. (And below)

cpp/basix/finite-element.h
```diff
@@ -749,6 +749,22 @@ class FiniteElement
   /// @return The dual matrix
   const xt::xtensor<double, 2>& dual_matrix() const;
 
+  /// Get the coefficients that define the polynomial set in terms of the
```
garth-wells, line 752: I think a bit more explanation with some maths would be helpful.

cpp/basix/finite-element.h
```diff
@@ -749,6 +749,22 @@ class FiniteElement
...
+  /// orthonormal polynomials.
+  ///
+  /// These coefficients are only stored for custom elements. This function will
+  /// throw an exception if called on a non-custom element
+  const xt::xtensor<double, 2>& wcoeffs() const;
+
+  /// Get the interpolation points for each subentity.
+  const std::array<std::vector<xt::xtensor<double, 2>>, 4>& x() const;
+
+  /// Get the interpolation matrices for each subentity.
```
garth-wells, line 762: Some math explanation would be helpful.

garth-wells (changes requested):

PR #410, opened 2022-04-06
Validate inputs to `create_custom_element`

### LINKED ISSUE #409: Check that information has the right shape when creating a custom element
Currently, passing incorrectly shaped data into a custom element causes a segfault. It would be better to detect the mistake and throw a runtime error

mscroggs:
Resolves #409

### DIFF
--- a/cpp/basix/math.cpp
+++ b/cpp/basix/math.cpp
@@ -86,3 +86,30 @@ basix::math::solve(const xt::xtensor<double, 2>& A, const xt::xarray<double>& B)
   return _B;
 }
 //------------------------------------------------------------------
+bool basix::math::is_singular(const xt::xtensor<double, 2>& A)
+{
+  assert(A.dimension() == 2);
+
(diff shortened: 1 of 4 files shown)


cpp/basix/math.cpp
```diff
@@ -86,3 +86,30 @@ basix::math::solve(const xt::xtensor<double, 2>& A, const xt::xarray<double>& B)
   return _B;
 }
 //------------------------------------------------------------------
+bool basix::math::is_singular(const xt::xtensor<double, 2>& A)
```
mscroggs, line 89: @IgorBaratta Is this a sensible way to check if a matrix is singular or should we be doing something else?

IgorBaratta: I think it's ok for now. 
Maybe we can reintroduce a function to check for singularity using `DGECON` (condition number) and `DGETRF` (LU factorization) later, since it's more precise but it can be more complicated.

PR #417, opened 2022-04-11
Add scaled identity map

mscroggs:
(This probably isn't of much use until after #405 is merged)


jpdean:
Looks good to me!

mscroggs:
Looks like this might be the "L2 Piola" map that ufl supports (and we might want to divide by detJ rather than multiply?)

https://github.com/FEniCS/ufl/blob/bcae96560f0a2c7b59b57397b155046d5846f116/ufl/algorithms/apply_function_pullbacks.py#L94-L96

jpdean:
Yes, it is. The mapped space is only L^2-conforming. I've also seen it called the "broken" Piola map. I believe that the pullback is defined as multiplication by det J. Obviously the pushforward is then division by det J.

Issue #418, opened 2022-04-14
Dot product in tabulate is inefficient


IgorBaratta:
When tabulate is called with thousands of points the `math::dot` is inefficient, and in this case we should offload the computation to a blas implemenation.

This computation large gemm is performed `[num derivative x value size]` times, as in https://github.com/FEniCS/basix/blob/main/cpp/basix/finite-element.cpp#L744.  

The index `p` in the loop is also a bit misleading, I first thought that we were computing `number of points` small matrix multiplications, but `p` refers to the number of derivatives.

mscroggs:
Can we now close this?

IgorBaratta:
Fixed in #453 .

Issue #395, opened 2022-03-11
Transpose polysets


mscroggs:
Many operations in polyset.cpp could be done on contiguous memory if the tables `P` were indexed `P(polynomial_index, derivative_index, point_index)`.

Currently the indexing is P(derivative_index, point_index, polynomial_index)`.

PR #425, opened 2022-04-19
Move ifs out of loops in tetrahedron polyset

### LINKED ISSUE #421: Move ifs out of loops in tetrahedron polyset

mscroggs:
Resolves #421.


IgorBaratta:
Not sure if this improves performance, but I think it will worsen code readability.
In the current implementation vectorization takes place inside `xtensor`, and I'm not sure if branch prediction is a big issue in outer loops.

mscroggs:
I think I'm going to abandon this for now. It was on on quads and hexes, but it makes the code so much worse for other cells. Perhaps this is worth revisiting after #395 and #418

PR #448, opened 2022-04-26
Speed up sympy tests

mscroggs:
Write the basis functions into each file rather than using sympy to compute them every time the test runs


mscroggs:
yes, I copied and pasted results from [symfem](https://github.com/mscroggs/symfem)

garth-wells (approved):

PR #453, opened 2022-04-27
Use blas for large matrix-matrix multiplications 

IgorBaratta:
```m * n * k > 16 * 16 * 16```

### CODE REVIEW

**Comment by garth-wells** [comment]
@IgorBaratta do you have timings on the performance improvement?

**Comment by IgorBaratta** [comment]
> @IgorBaratta do you have timings on the performance improvement?

I'll add some graphs 
For some elements/degrees polyset still dominates the total cost. But when the number of points is large it's always noticeable.

**Comment by IgorBaratta** [comment]
```c++
auto start = std::chrono::high_resolution_clock::now();

for (int i = 0; i < 10; i++)
  lagrange.tabulate(0, points, tab);

auto end = std::chrono::high_resolution_clock::now();
auto t = std::chrono::duration_cast<duration>(end - start);
```

## Tets:
tets

## Hexes
hexes

mscroggs (approved):

Issue #462, opened 2022-04-28
cblas.h not found on red hat


mscroggs:
See eg https://github.com/FEniCS/dolfinx/runs/6209738875?check_suite_focus=true

PR #463, opened 2022-04-28
Add Red Hat CI

mscroggs:
This will be useful to prevent issues like #462 in future


mscroggs:
I'll merge this but leave Red hat off the required tests for now as it currently fails

PR #468, opened 2022-04-28
add version updating utility

mscroggs:
Added a Python script that can be used to easily update all instances of the version number


chrisrichardson:
Looks good

PR #469, opened 2022-04-28
Define USE_ACCELERATE_BLAS in case apple accelerate library is being used

IgorBaratta:


garth-wells:
Not required following #489 change.

PR #479, opened 2022-04-29
Add maybe unused to variables in assert

jorgensd:


jorgensd:
> I think it's better to put the expression directly in the assertion.

Fixed

PR #488, opened 2022-05-06
Add CPATH to get wheel build working on RedHat

jhale:


garth-wells (changes requested):
The BLAS change in this PR will not be required with #489.

PR #495, opened 2022-05-13
Orthogonalise polynomial sets

mscroggs:
Orthogonalise the polynomial basis during element construction.

This reduces the condition number of the dual matrices. eg for serendipity elements (built-in = current elements on main, orthogonalised = after this PR):

image

This has a knock on effect of reducing the error when solving problems, eg:

image

### DIFF
--- a/cpp/basix/finite-element.cpp
+++ b/cpp/basix/finite-element.cpp
@@ -28,6 +28,7 @@ using namespace basix;
 namespace
 {
 //-----------------------------------------------------------------------------
+/// This function orthogonalises and normalises the rows of a matrix in place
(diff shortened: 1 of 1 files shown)


garth-wells (approved):

cpp/basix/finite-element.cpp
```diff
@@ -28,6 +28,20 @@ using namespace basix;
 namespace
 {
 //-----------------------------------------------------------------------------
+void orthogonalise(xt::xtensor<double, 2>& wcoeffs)
```
garth-wells, line 31: A doc comment would be helpful here, e.g. saying that rows of `wcoeffs` are orthogonalised in-place.

Also rows are normalised - would be good to say this.

PR #524, opened 2022-05-27
Transpose polynomial::tabulate

### LINKED ISSUE #514: Transpose `polynomials::tabulate`
To match the transposing of `polyset::tabulate`

mscroggs:
Fixes #514.

After this is merged, one of the DOLFINx demos needs updating: https://github.com/FEniCS/dolfinx/pull/2213


chrisrichardson:
Are we going to get rid of `polynomials::tabulate`, or is there a point in having other types (not Legendre)?

mscroggs:
I'm going to keep this function for now, as it looks like we might need Bernstein polynomials too. (We can always remove it in future if it turns out we don't need these)

PR #526, opened 2022-06-07
Attempt at automatic typing for basix

jhale:
This was created using mypy's automatic stub creator. I then replaced np.ndarray with npt.NDArray by automatic substitution.

In practice though it did not work in my nvim/pyright setup.

Can anyone else try it? It's not clear to me how this should be automated/updated either.


mscroggs:
I'm going to merge this, as having these types set is good and we can update manually for now. I've opened an issue for automating this: https://github.com/FEniCS/basix/issues/536

garth-wells:
@jhale what's the command for generating the stubs?

mscroggs:
`stubgen -m basix._basixcpp` I think

PR #560, opened 2022-07-12
Add `tensor product representation` properties to UFL element wrappers

IgorBaratta:
Used in `igor/tensor` branch of ffcx.


python/basix/ufl_wrapper.py
```diff
@@ -351,6 +360,17 @@ def is_custom_element(self) -> bool:
         """True if the element is a custom Basix element."""
         return self._is_custom
 
+    @property
+    def has_tensor_product_factorisation(self) -> bool:
+        """True if the element has a tensor product factorisation."""
+        return self.element.has_tensor_product_factorisation
+
```
garth-wells, line 367: It’s not clear from the docstring what is returned. Could some more explanation be added?

garth-wells: @mscroggs?

mscroggs: I've added a bit of detail to the docstring

mscroggs:
https://github.com/FEniCS/dolfinx/pull/2282 should fix the failing DOLFINx integration tests

PR #567, opened 2022-07-21
Only precompute matrices when they are not permutations

mscroggs:
When transformations are permutations, the precomputed matrices will be {permutation, identity, ones}. The permutation has already been computed, so we can save time by storing those rather than running `prepare_matrix`


IgorBaratta (approved):
lgtm!

garth-wells (approved):

PR #568, opened 2022-07-21
Use check on condition number rather than smallest eigenvalue

mscroggs:


cpp/basix/math.cpp
```diff
@@ -86,6 +94,34 @@ basix::math::eigh(const std::span<const double>& A, std::size_t n)
...
+double basix::math::rcond(const std::span<const double>& A, std::size_t n)
+{
+  // Copy A
+  std::vector<double> M(A.begin(), A.end());
+
+  int N = n;
+  char norm = '1'; // Compute the 1-norm
+  int ldA = n;
+  int info;
+  std::vector<double> work(4 * n);
```
garth-wells, line 106: It would be better to requires the caller to pass in these work buffers.

cpp/basix/math.h
```diff
@@ -68,6 +68,14 @@ std::array<typename U::value_type, 3> cross(const U& u, const V& v)
 std::pair<std::vector<double>, std::vector<double>>
 eigh(const std::span<const double>& A, std::size_t n);
 
+/// For a square matrix A, compute the norm of A multiplied by the reciprocal of
```
garth-wells, line 71:
```suggestion
/// For a square matrix A, compute the norm of A multiplied by the (estimated) reciprocal of
```

garth-wells, line 71: Which norm?

cpp/basix/precompute.cpp
```diff
@@ -40,6 +40,10 @@ precompute::prepare_matrix(
   mdarray2_t permuted_matrix(dim, dim);
   std::vector<T> diag(dim);
 
+  const double rcond_matrix = math::rcond(
```
garth-wells, line 43: Need a check here that the storage of `matrix` is contiguous, and check that it is square.

garth-wells (changes requested):
What speed-up does the change give?

mscroggs:
Closing this as #572 is a much better option

PR #573, opened 2022-07-25
Fix Sphinx version

mscroggs:
Sphinx 5.1.0 is broken


jhale:
@mscroggs Is this fixed in a later release?

https://www.sphinx-doc.org/en/master/changes.html

PR #585, opened 2022-08-18
Fix various typos

luzpaz:
Found via `codespell -q 3 -L ans,nce,nd`

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -42,7 +42,7 @@ pip install .
 
 ## Running the unit tests
 
-To install Basix and the extra depedencies required to run the Python unit tests:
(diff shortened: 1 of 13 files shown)


jhale:
Thanks @luzpaz! In addition, it would be nice to add this as a non-blocking CI step:

https://github.com/FEniCS/basix/blob/main/.github/workflows/pythonapp.yml#L58
https://github.com/FEniCS/basix/blob/main/setup.py#L26
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

luzpaz:
@jhale checkout https://github.com/codespell-project/actions-codespell

garth-wells (approved):

PR #586, opened 2022-08-22
Specify C++20 in wheel integration test CMakeLists.txt

jhale:
Basix requires C++20, so the `test_cmake/` CMakeLists.txt file should explicitly specify this.

I will also makes sure that this is run as part of the normal CI tests.


jhale:
I will backport this to make a `v0.5.0.post0` release.

PR #592, opened 2022-09-09
Add docs.h back into repo

### LINKED ISSUE #591: C++ core and Python bindings have conflicting Python dependency
Configuring the C++ portion of `basix` has an optional (and implicitly evaluated) Python dependency for generating documentation: https://github.com/FEniCS/basix/blob/main/cpp/CMakeLists.txt#L33-L40

The Python wrapper needs this documentation namespace to be generated in order to compile, e.g.: https://github.com/FEniCS/basix/blob/main/python/wrapper.cpp#L63

If the C++ component is compiled without a Python3 interpreter defined, building the Python layer is not possible.

This is problematic for systems which isolate build tools and particularly with Spack.

mscroggs:
Fixes #591

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ cpp/CMakeCache.txt
 cpp/CMakeFiles
 cpp/cmake_install.cmake
 cpp/basix/version.h
-cpp/basix/docs.h
 
 doc/python/build
 doc/python/source/generated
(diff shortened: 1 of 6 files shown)


garth-wells:
With DOLFIN, we tried over many years to 'auto-generate' Python docs from C++, but always failed.

The only approach that has worked reliably is to (1) minimise the number of functions exposed in Python, then (2) document these manually in a `.py` file.

garth-wells (approved):

PR #597, opened 2022-09-26
Make "CG" work and use identiy mapping for mixed elements whose subelements all use identity

mscroggs:


SarahRo (approved):
This fixes all issues I have encountered.

PR #599, opened 2022-09-26
Add deprecation warning when using "CG"

mscroggs:


mscroggs:
I think the timeline is dependent UFL / newFL: as long as UFL allows `"CG"` elements, this should probably stay

garth-wells (approved):

PR #609, opened 2022-10-17
Tidy up CI and packaging

jhale:
- Removes pip install from the CI (except for pytest-xdist in a few cases). If I could remind people *not* to pip install dependencies in the CI pipelines; there is always a better and more maintainable way to do this!
- Use `python3 -m` throughout, except on Mac, where it seems you must use `python -m` to get the the GitHub action activated versions of Python.
- Adds fenics-ufl as an optional dep installed from git.
- When building C++ demos, the cmake configuration file now asks Python for the location of the basix directory.
- One of the tests now uses the split C++/Python build as used in Spack.
- Moved to Redhat 9 compatible build where GCC 11 is default.
- Ensure that macOS builder uses the setup python installed python.
- Small updates to wheel builder.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -171,6 +171,7 @@ jobs:
           sudo cmake --install build-dir
 
       - name: Install Basix Python wrapper
+        run: |
           cd python
           python3 -m pip install .[test]
 
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -9,8 +9,7 @@ on:
     branches:
       - "**"
     tags:
-      - "*.*.*.*"
-      - "*.*.*"
+      - "v*"
   pull_request:
     branches:
       - main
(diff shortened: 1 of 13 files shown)


.github/workflows/pythonapp.yml
```diff
@@ -167,23 +161,24 @@ jobs:
...
+      - name: Install Basix C++ library
         run: |
-          pip install pytest sympy numba scipy matplotlib pytest-xdist pybind11
-          pip install git+https://github.com/FEniCS/ufl.git
-
-      - name: Install Basix
-        run: |
-          cmake -DCMAKE_BUILD_TYPE=Release -DDOWNLOAD_XTENSOR_LIBS=ON -B build-dir -S .
+          cd cpp
+          cmake -DCMAKE_BUILD_TYPE=Release -B build-dir -S . # Use make (not ninja)
```
jorgensd, line 169: Why do we not want to use `ninja`?

jhale: Good to do a build with the make generator, used frequently on HPCs (I know I don't bother installing ninja!). I left a comment there to note that it's not a mistake.

mscroggs (approved):

Issue #602, opened 2022-09-26
Add C-interface for some functionality


mscroggs:
We should allow users to call some functionality from C, including generating quadrature points and tabulating basis functions.

(@sclaus2 wants to use these inside kernels for CutFEM)

PR #612, opened 2022-10-20
Add C wrapper to basix for tabulation in C-code

sclaus2:
C-wrapper for basix functions tabulate_shape and tabulate. 
This C-wrapper is used to evaluate basis functions and their derivatives at a given set of points in generated C-code. This is needed for run-time quadrature evaluation used in methods such as CutFEM. This pull request contains the C-Wrapper as well as some small modifications of CMakeLists.txt in basix and is linked to #602 . @mscroggs @chrisrichardson


cpp/basix/cwrapper.cpp
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2022 Susanne Claus
```
jhale, line 1: We use clang-format on our C++ files.

cpp/basix/cwrapper.cpp
```diff
@@ -0,0 +1,61 @@
...
+extern "C" {
+
+basix_element* basix_element_create(const int basix_family, const int basix_cell_type, const int degree, 
+                                    const int lagrange_variant, const int dpc_variant, const bool discontinuous)
+{
+    //Specify which element is needed
+    basix::element::family family = static_cast<basix::element::family>(basix_family); 
+    basix::cell::type cell_type = static_cast<basix::cell::type>(basix_cell_type); 
+    int k = degree;
+    basix::element::lagrange_variant lvariant = static_cast<basix::element::lagrange_variant>(lagrange_variant);
```
jhale, line 21: `lvariant` etc. should also be `const`.

cpp/basix/cwrapper.h
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2022 Susanne Claus
```
jhale, line 1: Could you add an example of use in e.g. `demo/c/wrapper`?

That's often helpful for getting the API looking right.

cpp/basix/cwrapper.cpp
```diff
@@ -0,0 +1,61 @@
...
+// This file defines the C-API functions listed in cwrapper.h 
+// This is a C++ file that defines C calls 
+#include <basix/finite-element.h>
+#include <basix/cell.h>
+#include <span>
+#include "cwrapper.h"
+
+extern "C" {
+
+basix_element* basix_element_create(const int basix_family, const int basix_cell_type, const int degree,
```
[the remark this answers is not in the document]
jhale: A good alternative would be to have C enum equivalents of the existing C++ enums, as mentioned above.

cpp/basix/cwrapper.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "cwrapper.h"
+
+extern "C" {
+
+basix_element* basix_element_create(const int basix_family, const int basix_cell_type, const int degree, 
+                                    const int lagrange_variant, const int dpc_variant, const bool discontinuous)
+{
+    //Specify which element is needed
+    basix::element::family family = static_cast<basix::element::family>(basix_family); 
+    basix::cell::type cell_type = static_cast<basix::cell::type>(basix_cell_type);
```
jhale, line 19: `reinterpret_cast` might work?

jhale: Possibly not a good idea

garth-wells:
No activity for some time, so closing.

PR #617, opened 2022-11-09
Pass manifold cell information to base element of vector elements

jorgensd:
Possible fix for https://github.com/FEniCS/dolfinx/issues/2389

Prior to this fix, the ufl_cell() in the basix sub element would not be correct.


mscroggs:
looks like a new version of mypy may have come out. I'll look at fixing these errors

PR #629, opened 2022-12-21
Fix OneAPI tests

jhale:
Doesn't seem to find pip, so setup Python via GitHub action.


garth-wells:
This is not a correct fix. Intel oneAPI provides Python. The provided Numpy, etc, is consistently linked to MKL and other libraries. This avoids various hacks.

PR #631, opened 2022-12-22
Fix oneAPI tests

### LINKED ISSUE #628: OneAPI pipeline broken
Running the github workflow on main yields:
https://github.com/FEniCS/basix/actions/runs/3741962934/jobs/6352266601

garth-wells:
Fixes #628 and reverts #629.

The oneAPI basekit no longer includes Python.

### DIFF
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -1,6 +1,6 @@
-name: oneAPI compilers
+name: oneAPI
 
-# Test Basixusing Intel oneAPI compilers and MKL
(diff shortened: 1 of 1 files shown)


jhale:
Right you are! Thanks for the amendment. Where should libstdc++ be coming from?

garth-wells:
> Right you are! Thanks for the amendment. Where should libstdc++ be coming from?

The `intel-basekit` installs `libstdc++-11-dev`. Installing  `libstdc++-11-dev` seems to do the trick.

PR #636, opened 2023-02-21
Fix degree method in UFL wrapper layer for mixed elements

garth-wells:
Fixes https://github.com/FEniCS/dolfinx/issues/2543.


francesco-ballarin:
Thanks @garth-wells for the very quick fix, I confirm that this fixes the issues we had in our use case of mixed elements!

PR #638, opened 2023-02-25
Fix typo in basix.sobolev_spaces module name

francesco-ballarin:
(CI failure seems unrelated)


garth-wells:
I've added some tests, which exposed the bug #640.

garth-wells (approved):

PR #646, opened 2023-03-08
Simplify transformations

chrisrichardson:
* distil a lot of duplicated code
* use permutations instead of matrices where possible

### DIFF
--- a/cpp/basix/finite-element.h
+++ b/cpp/basix/finite-element.h
@@ -813,8 +813,7 @@ public:
   /// @param block_size The number of data points per DOF
   /// @param cell_info The permutation info for the cell
   template <typename T>
-  void apply_dof_transformation_to_transpose(const std::span<T>& data,
-                                             int block_size,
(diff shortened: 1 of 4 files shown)


cpp/basix/finite-element.h
```diff
@@ -1051,6 +1049,34 @@ class FiniteElement
   int interpolation_nderivs() const;
 
 private:
+  // Data permutation
+  // @param data Data to be permuted
+  // @param block_size
+  // @param cell_info Cell bitmap selecting required permutations
+  // @param eperm Permutation to use
+  // @param post Whether reflect is pre- or post- rotation.
+  template <typename T, bool post>
```
garth-wells, line 1058: C++20 `concepts` would be helpful for `post`.

mscroggs (approved):

mscroggs:
I thought I broke your PR, but the failing test was caused by https://github.com/FEniCS/basix/issues/651

PR #649, opened 2023-03-13
Allow UFL element to be created from UFL cell

### LINKED ISSUE #648: Allow UFL element to be created from UFL cell
Currently, there are a lot of `.cellname()`s needed when creating elements in the DOLFINx tests and demos, see https://github.com/FEniCS/dolfinx/pull/2381

mscroggs:
Fixes #648

### DIFF
--- a/python/basix/ufl_wrapper.py
+++ b/python/basix/ufl_wrapper.py
@@ -1212,7 +1212,8 @@ def _compute_signature(element: _basix.finite_element.FiniteElement) -> str:
 
 
 @_functools.lru_cache()
-def create_element(family: _typing.Union[_basix.ElementFamily, str], cell: _typing.Union[_basix.CellType, str],
+def create_element(family: _typing.Union[_basix.ElementFamily, str],
+                   cell: _typing.Union[_basix.CellType, str, _ufl.cell.Cell],
                    degree: int, lagrange_variant: _basix.LagrangeVariant = _basix.LagrangeVariant.unset,
                    dpc_variant: _basix.DPCVariant = _basix.DPCVariant.unset, discontinuous: bool = False,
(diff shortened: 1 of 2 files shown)


python/basix/ufl_wrapper.py
```diff
@@ -1229,6 +1230,8 @@ def create_element(family: _typing.Union[_basix.ElementFamily, str], cell: _typi
         gdim: The geometric dimension of the cell.
 
     """
+    if isinstance(cell, _ufl.cell.Cell):
```
garth-wells, line 1233: Could try/except be used here? Nice to avoid C++ style type checks if we can.

mscroggs: I can't see a way to do it that isn't a complete mess.

garth-wells: What if we only accept a `basix.CellType`? A lot of merit to there being only one way to do something.

mscroggs: `create_element` is a convenience function in the UFL wrapper to make it more user friendly to create elements. I've aimed to make the interface of this function as similar to the current UFL `FiniteElement` constructor as possible and it saves most users from having to write garbage like: 

```python
element = basix.ufl_wrapper.Basix_element(basix.create_element(basix.ElementFamily.P, basix.CellType.triangle, 2, basix.LagrangeVariant.equispaced))
```

mscroggs: But potentially accepting strings is enough and we don't need this additional option for input

mscroggs:
Closing this as https://github.com/FEniCS/dolfinx/pull/2604 is a better solution to this

PR #654, opened 2023-03-14
Allow permuted dof layout for Lagrange elements

chrisrichardson:
This should make it possible to support things like Sum Factorization and Bernstein.


garth-wells (commented):
This needs some really clear docs to explain what is going on.

chrisrichardson:
Yes.

PR #655, opened 2023-03-17
Rename ufl_wrapper -> ufl, and simplify functions in UFL wrapper

mscroggs:
Coupled with https://github.com/FEniCS/ffcx/pull/563 and https://github.com/FEniCS/dolfinx/pull/2381


garth-wells:
Could we reduced `finite_element`, `vector_element`, `tensor_element` to `element` and add a shape/dim argument? It's handy to have the dim match the spatial dim by default, but maybe there is a nice way to do this.

mscroggs:
We could. Currently I've gone for an interface similar to UFL to make switching easier for users

garth-wells:
> We could include the following keywords in the element function:
> 
> ```python
> element(..., vector=False, tensor=False, shape=None)
> ```
> 
> Then creating an element with `vector=True` would create a vector element of length tdim, but the shape argument could be used to change the size. Similarly with `tensor`.
> 
> However, I feel like this is more clunky than having three separate functions

We could have 
```python
element(..., rank=0, shape=None)
```
where `element(...) -> element(...) `, `vector_element(...) -> element(..., rank=1)`, `tensor_element(...) -> element(..., rank=2)`.

mscroggs:
> > We could include the following keywords in the element function:
> > ```python
> > element(..., vector=False, tensor=False, shape=None)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then creating an element with `vector=True` would create a vector element of length tdim, but the shape argument could be used to change the size. Similarly with `tensor`.
> > However, I feel like this is more clunky than having three separate functions
> 
> We could have
> 
> ```python
> element(..., rank=0, shape=None)
> ```
> 
> where `element(...) -> element(...) `, `vector_element(...) -> element(..., rank=1)`, `tensor_element(...) -> element(..., rank=2)`.

This is better than my attempt. I'm happy to go with this unless anyone else has objections or different ideas

jorgensd:
> > > We could include the following keywords in the element function:
> > > ```python
> > > element(..., vector=False, tensor=False, shape=None)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Then creating an element with `vector=True` would create a vector element of length tdim, but the shape argument could be used to change the size. Similarly with `tensor`.
> > > However, I feel like this is more clunky than having three separate functions
> > 
> > 
> > We could have
> > ```python
> > element(..., rank=0, shape=None)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where `element(...) -> element(...) `, `vector_element(...) -> element(..., rank=1)`, `tensor_element(...) -> element(..., rank=2)`.
> 
> This is better than my attempt. I'm happy to go with this unless anyone else has objections or different ideas

We would have to add a lot of extra documentation to explain this (in basix and in DOLFINx), as most people do not know what `rank` would signify.
Also we would have to add clear descriptions of what the default value is.

This kind of goes to opposite way of `assemble_scalar`, `assemble_vector`, `assemble_tensor`, as we could argue that we should just have `assemble(form, rank)`.

garth-wells:
> > > > We could include the following keywords in the element function:
> > > > ```python
> > > > element(..., vector=False, tensor=False, shape=None)
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Then creating an element with `vector=True` would create a vector element of length tdim, but the shape argument could be used to change the size. Similarly with `tensor`.
> > > > However, I feel like this is more clunky than having three separate functions
> > > 
> > > 
> > > We could have
> > > ```python
> > > element(..., rank=0, shape=None)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > where `element(...) -> element(...) `, `vector_element(...) -> element(..., rank=1)`, `tensor_element(...) -> element(..., rank=2)`.
> > 
> > 
> > This is better than my attempt. I'm happy to go with this unless anyone else has objections or different ideas
> 
> We would have to add a lot of extra documentation to explain this (in basix and in DOLFINx), as most people do not know what `rank` would signify. Also we would have to add clear descriptions of what the default value is.
> 
> This kind of goes to opposite way of `assemble_scalar`, `assemble_vector`, `assemble_matrix`, as we could argue that we should just have `assemble(form, rank)`.

No, because we can't uniquely determine the return type from the assembler arguments in all cases.

The current design lacks logic. Not sure how one makes sense of:
```python
e0 = element("Lagrange", ...)
e1 = vector_element("Lagrange", ...)
e2 = element("Raviart-Thomas" ...)
e3 = vector_element("Raviart-Thomas" ...)
e4 = tensor_element("Raviart-Thomas" ...)
```
With the proposal, one could iterate programmatically over all vector-valued elements just be element name e.g.,
```python
elements = ["Lagrange", "Raviart-Thomas", ...]
for name in elements:
    e = element(name, "triangle", rank=1)
```

python/basix/ufl.py
```diff
@@ -1236,23 +1214,42 @@ def _compute_signature(element: _basix.finite_element.FiniteElement) -> str:
...
-def create_element(family: _typing.Union[_basix.ElementFamily, str], cell: _typing.Union[_basix.CellType, str],
-                   degree: int, lagrange_variant: _basix.LagrangeVariant = _basix.LagrangeVariant.unset,
-                   dpc_variant: _basix.DPCVariant = _basix.DPCVariant.unset, discontinuous: bool = False,
-                   gdim: _typing.Optional[int] = None) -> BasixElement:
+def element(family: _typing.Union[_basix.ElementFamily, str], cell: _typing.Union[_basix.CellType, str], degree: int,
+            lagrange_variant: _basix.LagrangeVariant = _basix.LagrangeVariant.unset,
+            dpc_variant: _basix.DPCVariant = _basix.DPCVariant.unset, discontinuous: bool = False,
+            gdim: _typing.Optional[int] = None, rank: _typing.Optional[int] = None,
+            shape: _typing.Optional[_typing.Tuple[int, ...]] = None,
+            symmetry: _typing.Optional[bool] = None) -> BasixElement:
```
jorgensd, line 1222: Why does the element have to be a `BasixElement` if we are removing `ufl`? Could the element just be a `basix.ufl.FiniteElement`? Having it as a `basix.ufl.BasixElement` seems rather weird to me.

test/test_ufl_wrapper.py
```diff
@@ -42,9 +42,9 @@ def test_create_tensor_element(inputs):
     (basix.ElementFamily.P, "triangle", 2),
 ])
 def test_tensor_element_hash(inputs):
-    e = ufl_wrapper.create_element(*inputs)
-    sym = ufl_wrapper.TensorElement(e, symmetric=True)
-    asym = ufl_wrapper.TensorElement(e, symmetric=None)
+    e = basix.ufl.element(*inputs)
+    sym = basix.ufl.BlockedElement(e, shape=(2, 2), symmetry=True)
+    asym = basix.ufl.BlockedElement(e, shape=(2, 2), symmetry=False)
```
jorgensd, line 47: It seems a bit inconsistent to use `CamelCase` for `BlockedElement` but not for the creation of a `BasixElement`.

I am wondering, do we have a similar wrapper from `BlockedElement` called `blockedelement`? 
Having different kinds of casing for initialization of elements seems a bit confusing.

python/basix/ufl.py
mscroggs: I'm amazed that mypy was ok with this, this function can return a `BlockedElement` which is not a subclass of `BasixElement`.

test/test_ufl_wrapper.py
mscroggs: I don't like this either. It's because some of these are directly initing the class, others are using convenience functions.

I think it makes sense to add a couple more convenience functions and encourage people to not use the classes directly

mscroggs:
> "creates a UFL element"

I've update this to "UFL compatible element" as that should make it clearer what it's doing

mscroggs:
> > The function `element` is neither an action verb nor a constructor of a type `Element`, which makes it unclear what it actually does.
> 
> I'd be happy with `create_element` instead, although @garth-wells was keen on dropping the `create_`

I'm leaning towards keeping `element` and moving to having user functions called:

```
element, enriched_element, mixed_element, blocked_element
```

The only other consistent alternative I can see is:

```
create_element, create_enriched_element, create_mixed_element, create_blocked_element
```

which seems overly verbose

garth-wells (changes requested):

python/basix/ufl.py
```diff
@@ -1398,38 +1368,124 @@ def _create_enriched_element(elements: _typing.List[_BasixElementBase],
...
+            shape = tuple(tdim for _ in range(rank))
+    if rank is None:
+        rank = len(shape)
+    if rank != len(shape):
+        raise ValueError("Incompatible rank and shape.")
+
+    return BlockedElement(sub_element, shape=shape, symmetry=symmetry, gdim=gdim)
+
+
+def convert_ufl_element(ufl_element: _FiniteElementBase) -> _ElementBase:
```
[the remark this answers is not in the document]
mscroggs: short-term convenience. It's needed so that we can handle elements created UFL, so will exist until the UFL work gets far enough to make it no longer needer.

(It gives a deprecation warning to encourage users to stop using it)

python/basix/ufl.py
```diff
@@ -308,7 +308,7 @@ def basix_sobolev_space(self):
         pass
 
 
-class BasixElement(_BasixElementBase):
+class BasixElement(_ElementBase):
```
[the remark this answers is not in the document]
mscroggs: I've added to the docstrings

mscroggs:
I've added to the docstrings for all the element classes, and I've renamed the classes to have `_` at the start to suggest that they are internal and should not be used directly

garth-wells (approved):

PR #659, opened 2023-04-14
Use np.tile for blocked tabulation

michalhabera:
Fixes tabulation of BlockedElement to produce tables of shape `(num_points, num_sub_element_dofs * block_size)`, since current main returns `(num_points, num_sub_element_dofs * block_size**2)`.

Reproducible with
```python
import dolfinx
from mpi4py import MPI

mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 1, 1)
V = dolfinx.fem.VectorFunctionSpace(mesh, ("Lagrange", 1))
basis_wrapped = V.ufl_element().tabulate(0, [[0.0, 0.0]])
print(basis_wrapped.shape)
```
returns `(1, 1, 12)` instead of `(1, 1, 6)`.

### DIFF
--- a/python/basix/ufl.py
+++ b/python/basix/ufl.py
@@ -994,10 +994,8 @@ class _BlockedElement(_ElementBase):
         assert self.value_size == self._block_size  # TODO: remove this assumption
         output = []
         for table in self.sub_element.tabulate(nderivs, points):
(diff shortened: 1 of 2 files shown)


garth-wells:
@michalhabera could you add a test?

PR #661, opened 2023-04-17
Lazy element creation

mscroggs:
Make it so that element coefficients and DOF transformations are only created when they are needed.

This means that these computations do not need to be done in order to use a Basix element as a UFL element.


garth-wells (commented):
I think this ok, but having mutable state does have some downsides.

- Is this being done to hide an expensive operation that could be optimised to make it faster, or is it already optimised
- A downside or mutable state is that care is needed when using threading. Two thread calling `generate` will cause a race condition. Functions that could cause this need careful documentation.

Is there a practical was where we need a UFL element but don't eventually use the tabulation data?

Could you add to the docs which function(s) are now more expense at first call?

PR #668, opened 2023-05-02
Add reconstruct function to ufl wrapper

### LINKED ISSUE #667: Basix FiniteElement reconstruction
In some cases, it is very neat to be able to reconstruct an element, such as:
```python
 ufl_el = uh.function_space.ufl_element()
 new_el = ufl_el.reconstruct(degree=degree_raise+ufl_el.degree())
```
which uses https://github.com/FEniCS/ufl/blob/6eeef938c877634dcb7c8acb101bb2363ed1455b/ufl/finiteelement/finiteelement.py#L204-L217

Having a similar interface in basix would be neat, as finding the rank, lagrange-variant etc from the original element  is quite some work. i.e. having something like
```python
el = basix.ufl.element(......)
new_el = el.reconstruct_element(**kwargs)
```
where kwargs overload entries in the original constructor.

mscroggs:
Resolves #667.

See also https://github.com/FEniCS/ffcx/pull/569.


jorgensd:
@mscroggs  do we need to modify this now?

mscroggs:
Closing this as it needs rethinking following major UFL changes

PR #670, opened 2023-05-05
Relax some floating point comparisons

garth-wells:
Some checks were too tight for `float32`. Issues picked up in testing `float` support in DOLFINx.

Eventually more tested is required, but only `float64` is supported via the Python interface at the moment. Test should be added once the Python interface is extended.

### DIFF
--- a/cpp/basix/finite-element.cpp
+++ b/cpp/basix/finite-element.cpp
@@ -23,8 +23,6 @@
 #include <limits>
 #include <numeric>
 
-#include <iostream>
-
 #define str_macro(X) #X
 #define str(X) str_macro(X)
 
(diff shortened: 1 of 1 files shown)


cpp/basix/finite-element.cpp
```diff
@@ -952,7 +948,8 @@ FiniteElement<F>::FiniteElement(
     for (std::size_t col = 0; col < matM.extent(1); ++col)
     {
       F v = col == row ? 1.0 : 0.0;
-      if (std::abs(matM(row, col) - v) > 1.0e-12)
+      constexpr F eps = 100 * std::numeric_limits<F>::epsilon();
```
mscroggs, line 951: Should this be `10.0 * ` and `<float>` to match the other line? Or are these intentionally different?

garth-wells: Deliberate, but with without good justification ;).

I made the previous 'looser' because it follows a LU factorisation. I thought this check could be stricter. Proper 'fix' would be to support `float32` from Python and run the test suite, but we're not there yet.

mscroggs (approved):

PR #673, opened 2023-06-06
Test scikit-build-core and pyproject.toml for top-level install

chrisrichardson:


.github/workflows/pythonapp.yml
```diff
@@ -43,9 +43,10 @@ jobs:
           flake8 --statistics python/
           flake8 --statistics demo/python
 
-      - name: Run mypy checks
-        run: |
-          python3 -m mypy python/basix
+#      - name: Run mypy checks
```
garth-wells, line 46: Uncomment?

pyproject.toml
```diff
@@ -1,7 +1,24 @@
+[project]
+name = "fenics-basix"
+packages = ["basix"]
+version = "0.7.0.dev0"
+dependencies = ["numpy>=1.21"]
+
+[project.optional-dependencies]
+ci = ["pytest-xdist", "pytest", "numba", "scipy>=0.16", "flake8", "pydocstyle", "mypy",
```
garth-wells, line 8: Are all of these required? `pylit`?

Can we break the list up? Some are for running Basix tests, e.g. pytest, and other 'test' Basix, e.g. `flake8`.

setup.py.old
garth-wells, line 1: Remove file.

garth-wells (changes requested):

.github/workflows/pythonapp.yml
chrisrichardson: mypy seems to be out of control - I don't know how to get it to pass the tests - are there some options I need to set? Or does it need to use a different version?

pyproject.toml
chrisrichardson: I added these incrementally to get it to pass the CI, which does need `pylit`.
I guess we can group them, e.g. `ci` `test` `lint` etc. then we have to do `pip install.[ci, test, lint]`

PR #682, opened 2023-08-14
Move QuadratureElement and RealElement from ffcx to Basix

mscroggs:
This is the first part of https://github.com/FEniCS/ffcx/issues/590


jhale:
The .degree is confusing, it seems to be related to the number of points in the rule rather than the degree of the rule?

mscroggs:
> The .degree is confusing, it seems to be related to the number of points in the rule rather than the degree of the rule?

Good point. I've changed it to use the degree if provided, and default to number of points if the degree is not input

PR #696, opened 2023-09-14
Add a [ufl] install option to install Basix and UFL

mscroggs:
Now that `basix.ufl` is an intended way to create UFL elements, it might be useful to be able to run `pip install git+http://github.com/FEniCS/basix[ufl]` to install UFL with Basix.


garth-wells (commented):
This will another line that we'll need to update at each release - is it worth it?

mscroggs:
> This will another line that we'll need to update at each release - is it worth it?

I'm not sure. Currently, just installing it with optional is fine I guess as that only has numba on top of ufl. Maybe worth revisiting in future if we ever add a lot more stuff to optional

PR #708, opened 2023-09-28
Remove rank input to element and blocked_element

mscroggs:
This can be merged after https://github.com/FEniCS/ffcx/pull/618 and https://github.com/FEniCS/dolfinx/pull/2789


garth-wells (approved):
Much simpler!

PR #709, opened 2023-10-05
Update wheel build action.

jhale:


garth-wells (approved):
@jhale is this ready, or is it handled in the pyproject PR?

jhale:
It can go in. It's possible the minimum macOS target line should go into nanobind, but it's necessary for now.

PR #715, opened 2023-10-18
Remove `setup.py`

garth-wells:
Use `pyproject.toml` and `scikit-build-core`.


garth-wells:
@jhale , @mscroggs what do you think of this?

mscroggs (approved):

PR #716, opened 2023-10-23
Use scikit-build-core to handle minimums.

jhale:


python/pyproject.toml
```diff
@@ -21,4 +23,6 @@ test = ["pytest", "sympy", "scipy", "matplotlib", "fenics-basix[optional]"]
 ci = ["mypy", "pytest-xdist", "fenics-basix[docs,lint,test,optional]"]
 
 [tool.scikit-build]
-wheel.packages = ["basix"]
+minimum-version = "0.5"
+cmake.minimum-version = "3.19.0"
+wheel.packages = ["python/basix"]
```
garth-wells, line 28: This looks wrong to me - the `pyproject.toml` file is already in `python`/.

PR #720, opened 2023-10-27
Update install docs

garth-wells:
Includes instruction for editable Python install.

Basic instructions. More elaborations can be proposed later if helpful.

### DIFF
--- a/python/basix/__init__.py
+++ b/python/basix/__init__.py
@@ -2,9 +2,7 @@
 
 The core of the library is written in C++, but the majority of Basix's
 functionality can be used via this Python interface.
-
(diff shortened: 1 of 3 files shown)


jorgensd:
As this is not sufficient, I suggest we close this in favor of #721

python/basix/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 The core of the library is written in C++, but the majority of Basix's
 functionality can be used via this Python interface.
+
```
mscroggs, line 5: Is this newline an intentional addition?

PR #748, opened 2023-11-26
Use ruff for linting

garth-wells:


python/basix/_basixcpp.pyi
```diff
@@ -1,10 +1,11 @@
 from typing import Any, ClassVar, List, Tuple, overload
 
-import numpy
-import numpy.typing as npt
+import np.typing as npt
```
francesco-ballarin, line 3: The ordering here seems odd

python/basix/ufl.py
```diff
@@ -8,9 +8,10 @@
 from abc import abstractproperty as _abstractproperty
 from warnings import warn as _warn
 
-import numpy as _np
+import numpy as np
 import numpy.typing as _npt
```
francesco-ballarin, line 12: May want to import as `npt` is `numpy` is imported as `np` rather than `_npt`

python/basix/ufl.py
```diff
@@ -2011,12 +2231,14 @@ def custom_element(cell_type: _basix.CellType, value_shape: _typing.Union[_typin
         discontinuous,
         embedded_subdegree,
         embedded_superdegree,
-        polyset_type
+        polyset_type,
```
francesco-ballarin, line 2234: Unsure on why the extra comma was added

python/basix/ufl.py
```diff
@@ -2036,13 +2258,15 @@ def mixed_element(elements: _typing.List[_ElementBase], gdim: _typing.Optional[i
...
-                       weights: _typing.Optional[_npt.NDArray[_np.float64]] = None,
-                       pullback: _AbstractPullback = _ufl.identity_pullback) -> _ElementBase:
+def quadrature_element(
+    cell: _typing.Union[str, _basix.CellType],
+    value_shape: _typing.Tuple[int, ...] = (),
+    scheme: _typing.Optional[str] = None,
+    degree: _typing.Optional[int] = None,
+    points: _typing.Optional[_npt.NDArray[np.float64]] = None,
+    weights: _typing.Optional[_npt.NDArray[np.float64]] = None,
+    pullback: _AbstractPullback = _ufl.identity_pullback,
```
francesco-ballarin, line 2268: Also here extra comma added by ruff

python/pyproject.toml
```diff
@@ -18,7 +18,7 @@ dependencies = ["numpy>=1.21"]
 
 [project.optional-dependencies]
 docs = ["markdown", "pylit3", "pyyaml", "sphinx==5.0.2", "sphinx_rtd_theme"]
-lint = ["flake8", "pydocstyle", "isort"]
```
francesco-ballarin, line 21: I would advise against dropping isort altogether. See https://github.com/astral-sh/ruff/issues/6190

python/pyproject.toml
```diff
@@ -18,7 +18,7 @@ dependencies = ["numpy>=1.21"]
...
+lint = ["ruff"]
 optional = ["numba", "fenics-ufl@git+https://github.com/fenics/ufl"]
 test = ["pytest", "sympy", "scipy", "matplotlib", "fenics-basix[optional]"]
 ci = ["mypy", "pytest-xdist", "fenics-basix[docs,lint,test,optional]"]
```
francesco-ballarin, line 24: This is missing something like
```
[tool.ruff]
line-length = 120
```
to restore the previous behavior of
```
[flake8]
max-line-length = 120
```

test/test_cell.py
```diff
@@ -2,11 +2,10 @@
 # FEniCS Project
 # SPDX-License-Identifier: MIT
 
+import basix
```
francesco-ballarin, line 5: This (and same in all tests) seems odd. `basix` is the local library name, and `isort` would surely put it back at the bottom of the imports.

python/pyproject.toml
francesco-ballarin: Sorry, it's already in the `pyproject.toml` in the root folder.

pyproject.toml
```diff
@@ -68,11 +68,35 @@ archs = [
...
+    "I",   # isort
+    "RUF", # Ruff-specific rules
+    "UP",  # pyupgrade
+    # "ANN",
+    "ICN",
+    # "SIM",
+    # "PL",
+]
+ignore = []
+
```
francesco-ballarin, line 97: Maybe add
```
[tool.ruff.pycodestyle]
max-doc-length = 120
```
too?

Seems to be off by default
https://docs.astral.sh/ruff/settings/#pycodestyle-max-doc-length

pyproject.toml
```diff
@@ -68,11 +68,35 @@ archs = [
...
 ignore_missing_imports = true
 
-[tool.isort]
-line_length = 120
+[tool.black]
+line-length = 120
+
+[tool.ruff]
+line-length = 120
+indent-width = 4
```
francesco-ballarin, line 79: Probably unnecessary, it is 4 by default, see https://docs.astral.sh/ruff/settings/#indent-width

jhale:
Closed in favour of lighter approach in #774

PR #751, opened 2023-11-29
Do not store `.doctrees` binary files on website

francesco-ballarin:
Side effect of #744, since these files used to be stored in `build/doctrees` rather than `build/html/.doctrees`.

See e.g. https://github.com/FEniCS/docs/commit/9dbb9f04823216c6b659efc35a2fea10ede24818 for a diff

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -65,6 +65,7 @@ jobs:
           doxygen
           cd ../python
           python -m sphinx -W -b html source build/html
+          rm -rf build/html/.doctrees
       - name: Upload C++ documentation artifact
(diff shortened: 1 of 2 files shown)


.github/workflows/pythonapp.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           doxygen
           cd ../python
           python -m sphinx -W -b html source build/html
+          rm -rf build/html/.doctrees
```
francesco-ballarin, line 68: My understanding is that the files that get actually uploaded are those generated with `doc/web/make_html.py`, right? If so, the addition of this line is probably unnecessary, the addition that matters is the one in `doc/web/make_html.py`.

PR #767, opened 2024-01-08
Wrap functionality in Python with nicer typing and docs

mscroggs:
- Added python classes and functions that call functions in wrapper, rather than calling wrapper directly
- Added documentation and typing
- Removed `*_to_string` functions for enums, as `.name` can be used instead
- Removed semi-automated copying of docstrings from c++ to Python 🥳


garth-wells (approved):
Nice!

PR #780, opened 2024-02-02
Add create_tp_element function

mscroggs:
This function will create an Lagrange element on a quad or hex that uses tensor product ordering

Coupled with https://github.com/FEniCS/ffcx/pull/664


cpp/basix/finite-element.cpp
```diff
@@ -322,6 +322,189 @@ basix::create_element(element::family, cell::type, int,
...
+std::vector<int> basix::tp_dof_ordering(element::family family, cell::type cell,
+                                        int degree, element::lagrange_variant,
+                                        element::dpc_variant, bool)
+{
+  std::vector<int> dof_ordering;
+  std::vector<int> perm;
+
+  switch (family)
+  {
+  case element::family::P:
```
chrisrichardson, line 403: Should this throw an error for other families? Currently just returning `{}` I think?

mscroggs: Yes, throwing is probably better than returning an empty list

IgorBaratta (approved):

PR #782, opened 2024-02-02
Add a hash functions to finite elements

mscroggs:
Could be written into generated kernels then used to check that elements are the same.


garth-wells (commented):
Could we use `std::hash`?

https://en.cppreference.com/w/cpp/utility/hash

mscroggs:
> Could we use `std::hash`?
> 
> https://en.cppreference.com/w/cpp/utility/hash

That would probably be better

Issue #679, opened 2023-08-04
Replace Gram-Schmidt orthogonalisation with better algorithm


mscroggs:
When elements are created, the polynomial basis is orthogonalised:

https://github.com/FEniCS/basix/blob/0fc130f7d790f793ffa57bf8d056232ff6f12be8/cpp/basix/finite-element.cpp#L41-L62

This orthogonalisation currently uses the Gram-Schmidt process. Instead, we should use a better algorithm (eg. Computing Q from the QR decomposition of A^T)

ampdes:
The above function is now in [math.h](https://github.com/FEniCS/basix/blob/44962386b6a5f06f64ec956cdd5737984df81f32/cpp/basix/math.h#L362).

The Gram-Schmidt process constructs the Q matrix in-place whereas the R matrix can be constructed by collecting the `a` s (the `a` terms in above code snippet).

One alternative is the Householder reflection algorithm which constructs the R matrix in-place which is not sought here.

However, the current implementation is Classical Gram-Schmidt
https://github.com/FEniCS/basix/blob/44962386b6a5f06f64ec956cdd5737984df81f32/cpp/basix/math.h#L368C1-L391C1

can easily be changed to the numerically stable modified Gram-Schmidt:

```
  for (std::size_t i = start; i < wcoeffs.extent(0); ++i)
  {
    T norm = 0;
    for (std::size_t k = 0; k < wcoeffs.extent(1); ++k)
      norm += wcoeffs(i, k) * wcoeffs(i, k);
    if (std::abs(norm) < 4 * std::numeric_limits<T>::epsilon())
    {
      throw std::runtime_error(
          "Cannot orthogonalise the rows of a matrix with incomplete row rank");
    }

    for (std::size_t k = 0; k < wcoeffs.extent(1); ++k)
      wcoeffs(i, k) /= std::sqrt(norm);

    for (std::size_t j = i+1; j < wcoeffs.extent(1); ++j)
    {
      T a = 0;
      for (std::size_t k = 0; k < wcoeffs.extent(1); ++k)
        a += wcoeffs(i, k) * wcoeffs(j, k);
      for (std::size_t k = 0; k < wcoeffs.extent(1); ++k)
        wcoeffs(j, k) -= a * wcoeffs(i, k);
    }

  }
```

mscroggs:
[...]

PR #785, opened 2024-02-09
Change to the Modified Gram Schmidt algorithm in orthogonalise (#679)

ampdes:
* change the orthogonalization algorithm to modified gram schmidt

* norm is precomputed

* first check row rank before orthogonalizing the rows


jhale:
Looks good; please run `clang-format -i` on the file.

PR #787, opened 2024-02-10
Improve docstrings and add copyright statements

garth-wells:


python/basix/__init__.py
```diff
@@ -1,4 +1,9 @@
-"""Basix is a finite element definition and tabulation runtime library.
+# Copyright (C) 2023-204  Matthew Scroggs and Garth N. Wells
```
mscroggs, line 1:
```suggestion
# Copyright (C) 2023-2024  Matthew Scroggs and Garth N. Wells
```

Looks like this typo is in every file

PR #796, opened 2024-03-21
dtype when creating elements

mscroggs:
- Default dtype updated to match DOLFINx (if installed)
- dtype input added to custom element

Will allow us to do https://github.com/FEniCS/dolfinx/pull/3136


jhale:
What's the status of this?

garth-wells:
I'm not wild about bringing DOLFINx into this. I'd favour defaulting to `float64` or being completely explicit on the type.

jorgensd:
> I'm not wild about bringing DOLFINx into this. I'd favour defaulting to `float64` or being completely explicit on the type.

I agree with Garth

mscroggs:
> What's the status of this?

Needs a bit of work before it's ready. Should have time to do this work in the next couple of weeks

mscroggs:
> > What's the status of this?
> 
> Needs a bit of work before it's ready. Should have time to do this work in the next couple of weeks

It was less work to fix that I thought it might be. Ready for review now

python/basix/finite_element.py
```diff
@@ -582,7 +587,7 @@ def create_element(
     dpc_variant: DPCVariant = DPCVariant.unset,
     discontinuous: bool = False,
     dof_ordering: typing.Optional[list[int]] = None,
-    dtype: npt.DTypeLike = np.float64,
+    dtype: typing.Optional[npt.DTypeLike] = None,
```
garth-wells, line 590:
```suggestion
    dtype: npt.DTypeLike =  np.float64,
```

python/basix/finite_element.py
```diff
@@ -602,19 +607,24 @@ def create_element(
...
-            family.value,
-            celltype.value,
-            degree,
-            lagrange_variant.value,
-            dpc_variant.value,
-            discontinuous,
-            dof_ordering if dof_ordering is not None else [],
-            np.dtype(dtype).char,
-        )
+    if dtype is None:
```
garth-wells, line 610: Remove?

python/basix/finite_element.py
```diff
@@ -657,26 +668,56 @@ def create_custom_element(
...
+
+        dtype = default_scalar_type
+
+    if wcoeffs.dtype != dtype:
+        wcoeffs = np.dtype(dtype).type(wcoeffs)  # type: ignore
+        x = [[np.dtype(dtype).type(j) for j in i] for i in x]  # type: ignore
+        M = [[np.dtype(dtype).type(j) for j in i] for i in M]  # type: ignore
+    if np.issubdtype(dtype, np.float32):
+        return FiniteElement(
+            _create_custom_element_float32(
```
garth-wells, line 687: I wonder if using `std::variant` in the nanobind layer could avoid the duplication with  `_create_custom_element_float32` and  `_create_custom_element_float64`.

python/basix/finite_element.py
mscroggs: Doing it the way we currently do it allows us to do this in DOLFINx to make Basix use the same default scalar type, which is convenient:

https://github.com/FEniCS/dolfinx/blob/5332dc9f4b97054a63264d8fa4711ed5e5ff0b1a/python/dolfinx/__init__.py#L39-L41

python/basix/finite_element.py
garth-wells: @mscroggs

mscroggs: This is needed to allow DOLFINx to change the default scalar type as in the comment above

garth-wells: This all seems too implicit to me. I think DOLFINx should explicitly pass the type to Basix functions. Global defaults are not nice.

mscroggs: ok, default_scalar_type is gone

garth-wells (approved):

PR #799, opened 2024-04-13
Update `nb::ndarray` constructor in preparation of next nanobind release

francesco-ballarin:
Due to https://github.com/wjakob/nanobind/commit/937a1df52428f04a0a240cc5827fb3c6769cf5c2 which dropped a default value

Not supposed to be merged until the next nanobind release. Coupled with https://github.com/FEniCS/dolfinx/pull/3105

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -36,6 +36,14 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y doxygen graphviz libopenblas-dev liblapack-dev ninja-build
+      - name: Install nanobind from source
+        run: |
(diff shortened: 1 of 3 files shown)


francesco-ballarin:
To clarify: in contrast to the nanobind change discussed in the first post of https://github.com/FEniCS/dolfinx/pull/3105, the nanobind change addressed here is backward compatible: in principle, the PR could be marked as ready to review (and not draft), but I won't do that because the change introduced here isn't necessary for the current nanobind release.

garth-wells:
@francesco-ballarin I think we should merge this - it will keep the Basix working when nanobind 2.0 is released.

garth-wells (approved):

PR #802, opened 2024-04-18
Revise transformation names and add docs

garth-wells:
Add mathematical descriptions of the different degree-of-freedom transformation functions, and rename functions to more closely reflect the mathematical descriptions.

Numerous other documentation improvements.


cpp/basix/precompute.h
```diff
@@ -285,25 +279,26 @@ void pre_apply_matrix(
 
 /// @brief Apply a (precomputed) matrix to some transposed data.
 ///
+/// Computes \f$v = u M^{T}\f$.
```
garth-wells, line 282: @mscroggs is this doc line correct?

mscroggs: doesn't look right
```suggestion
/// Computes \f$v = M u\f$.
```

mscroggs: Yes, but maybe it's clearer to write this:
```suggestion
/// Computes \f$v^{T} = Mu^{T}\f$ (or equivalently \f$v = u M^{T}\f$).
```

mscroggs (approved):

PR #806, opened 2024-04-19
Add permute_subentity_closure function

mscroggs:


cpp/basix/finite-element.cpp
```diff
@@ -1133,6 +1133,259 @@ FiniteElement<F>::FiniteElement(
...
+      for (int i : dofs[0][0])
+        ref.push_back(i);
+      // Edges
+      for (int i : dofs[1][0])
+        ref.push_back(i);
+
+      if (!_dof_transformations_are_identity)
+      {
+        auto& trans1 = _eperm.at(cell::type::interval)[0];
+        if (dofs[1][0].size() > 0)
```
garth-wells, line 1173:
```suggestion
        if (!dofs[1][0].empty())
```

cpp/basix/finite-element.cpp
```diff
@@ -1133,6 +1133,259 @@ FiniteElement<F>::FiniteElement(
...
+      const auto conn = cell::sub_entity_connectivity(_cell_type)[1][0];
+      std::vector<std::vector<std::vector<int>>> dofs;
+      dofs.resize(2);
+      for (int dim = 0; dim <= 1; ++dim)
+      {
+        dofs[dim].resize(conn[dim].size());
+        for (std::size_t i = 0; i < conn[dim].size(); ++i)
+        {
+          const int e = conn[dim][i];
+          for (int j : _edofs[dim][e])
```
garth-wells, line 1153: Simpler would be:
```cpp
for (std::size_t j =0; j < _edofs[dim][e].size(); ++j)
   dofs[dim][i].push_back(dof_n++);
```

cpp/basix/finite-element.cpp
```diff
@@ -1133,6 +1133,259 @@ FiniteElement<F>::FiniteElement(
...
+          {
+            std::ignore = j;
+            dofs[dim][i].push_back(dof_n++);
+          }
+        }
+      }
+
+      std::vector<std::size_t> ref;
+      for (int i : dofs[0][1])
```
garth-wells, line 1162:
```cpp
ref.insert(ref.end(), dofs[0][1].begin(), dofs[0][1].end());
```

cpp/basix/finite-element.h
```diff
@@ -837,6 +837,64 @@ class FiniteElement
...
+  ///
+  /// @note This function is designed to be called at runtime, so its
+  /// performance is critical.
+  ///
+  /// @param[in,out] d Indices associated with each reference element
+  /// degree-of-freedom (in). Indices associated with each physical
+  /// element degree-of-freedom (out).
+  /// @param cell_info Permutation info for the cell
+  /// @param entity_type The cell type of the sub-entity
+  void permute_subentity_closure(std::span<std::int32_t> d,
```
jpdean, line 854: It looks like `cell_info` is really the information for the sub-entity (e.g. if the cell was a tetrahedron and the sub-entity was an interval, you would have to pass data for the interval, which would be either 0 or 1 depending on if it is flipped or not). Do you think we could instead make `cell_info` the actual cell information (e.g. containing the information for all sub-entities of the tet, as is returned by `get_cell_permutation_info` in DOLFINx) and also pass a sub-entity index to identify which cell sub-entity. I think this might be easier than the user having to figure out the info for a particular sub-entity

jpdean (commented):
Thanks for looking at this, Matt! I have one comment about the interface

cpp/basix/finite-element.h
mscroggs: I've added `permute_subentity_closure_inv` that goes the other way and should be what you want and I've added an overloaded version of that takes `cell_into` and the entity index (as well as renaming the input in the original version to `entity_info`)

PR #819, opened 2024-04-28
Windows build

jhale:
Uses vcpkg and currently the split build for debugging. Remaining error related to finding `_basixcpp` library on Python `import`. Needs tidying (passing lots of options/environment variables around at the moment). Also would be good to try combined build.


jhale:
This works except for C++ demos and proper re-integration with existing platforms and removing hard coded paths.

Split build is the 'proper' way to go on Windows - also how it will be done in conda.

jhale:
@minrk Any idea how to deal elegantly with this on Windows? https://github.com/FEniCS/basix/pull/819/files#diff-244eaae986544ece8f0043ffba8a5ceebed7d359074d684daecce7b7d3ce3328R14

pyproject.toml
```diff
@@ -34,6 +34,8 @@ testpaths = ["test"]
 minimum-version = "0.5"
 cmake.minimum-version = "3.19.0"
 wheel.packages = ["python/basix"]
+cmake.verbose = true
+logging.level = "DEBUG"
```
mscroggs, line 38: Do these lines need removing before merge?

python/basix/__init__.py
```diff
@@ -9,6 +9,13 @@
 functionality can be used via this Python interface.
 """
 
+import sys
+
+if sys.platform.startswith("win32"):
+    import os
+
+    os.add_dll_directory("D:/a/basix/install/bin")
```
mscroggs, line 17: Will this always be the correct install path?

pyproject.toml
jhale: Yes.

python/basix/__init__.py
jhale: No - I'm going to wait for @minrk comment on how to deal with this more elegantly. This is 'expected behaviour' on Windows, see e.g. https://docs.python.org/3/library/os.html#os.add_dll_directory

jhale:
This now passes all tests on Windows with Visual Studio - there is a remaining issue with how to deal with `os.add_dll_directory("D:/a/basix/install/bin")` which I am unsure about.

The ugly `if` around the cmake version and install is currently necessary due to RedHat's old cmake version - the Windows install uses features introduced in CMake 3.21.

python/basix/__init__.py
minrk: I don't know about elegant, but DLLs are generally searched for on $PATH (there is no LD_LIBRARY_PATH for Windows, just PATH).

As a result, the [typical installation directory for DLLs](https://github.com/zeromq/libzmq/blob/aa885c5a154256612108636b0fb22f44ae0e247a/CMakeLists.txt#L1693) on Windows is with

```
RUNTIME_DESTINATION ${CMAKE_INSTALL_BINDIR}
```

So for CI, you can either add the DLL location to %PATH%, or set the install location to somewhere already on %PATH%.

minrk: More references:

- conda build [docs on Windows shared libraries](https://docs.conda.io/projects/conda-build/en/stable/resources/use-shared-libraries.html#shared-libraries-in-windows), and
- conda-forge [example cmake build script](https://conda-forge.org/docs/maintainer/knowledge_base/#simple-cmake-based-bldbat).

jhale: I agree, but Python's object loader ignores the rule "DLLs are generally searched for on $PATH", see: https://docs.python.org/3/library/os.html#os.add_dll_directory

minrk: Right, I forgot about that whole mess. Conda-forge [patches](https://github.com/conda-forge/python-feedstock/blob/46e4991703e49cf75833699d42d6605147d31aa6/recipe/patches/0007-Add-CondaEcosystemModifyDllSearchPath.patch) Python to search in standard locations DLLs can install to. Python doesn't appear to have any of these by default except the system itself and the Extension DLL parent, at least according to the docs.

I guess the choices are:

1. record libbasix.dll location somewhere during Python build with e.g. `configure_file`, so there's a variable to import and call add_dll_dir on
3. get it from an env variable at runtime
3. install libbasix to one of the 'standard' locations, i.e. system directory
4. copy DLLs into the Python package ([delvewheel](https://pypi.org/project/delvewheel/) takes care of this post-build, like `delocate` or `auditwheel` for mac and linux)

If you want `pip install` to work by default without extra args, I think the only option is to copy libbasix into the Python package on Windows, then you only need `add_dll_directory(Path(__file__).parent)`. While delvewheel will do this, if you want users to get it without an extra step, you'll have to do the copy_file yourself.

jhale:
Summary:

- No hard coded paths in our source.
- Combined install works _without_ adhoc modifications to $PATH or Python's dll loader's internal concept of the PATH. All dlls (basix, openblas etc). end up in the same directory as the `.pyd` wrapper. Python's dll loader will search `.` for all runtime links, and it just works.
- Split install. vcpkg-provided dlls are put with the C++ basix library. The Python wrapper sits on its own. C++ runtime linking only succeeds if the C++ libraries are put on the PATH (standard). Python runtime linking only succeeds if and the Python DLL loader is told about the C++ runtime directory, which is consistent with the documentation (https://docs.python.org/3/library/os.html#os.add_dll_directory). This is currently done through a `conftest.py` file.
- No changes for Unix targets.
 
The nasty IFs can be removed once RedHat has a newer CMake (>3.21).

This could be tried in conda now @minrk.

jhale:
I've merged this with the idea that it will certainly need another iteration once tested in Conda.

PR #822, opened 2024-05-02
Windows: make runtime dependency install optional

minrk:
Bundling dependencies is not desirable in a conda package, where packaging already handles this, plus it happens to be throwing an error because blas/lapack DLLs aren't found by cmake (their DIRECTORIES aren't specified).

I believe this was the only change I needed to get Windows builds to work on conda-forge: https://github.com/conda-forge/fenics-basix-feedstock/pull/18

I'm not sure what the default should be, but I'd _guess_ it should be OFF by default. I left the default behavior unchanged in this PR.

FWIW, the default vs2019 compiler failed with:

```
D:\bld\fenics-basix-meta_1714546306829\work\cpp\basix\mdspan.hpp(3847,15): error C2672: 'std::experimental::detail::get_actual_static_padding_value': no matching overloaded function found [D:\bld\fenics-basix-meta_1714546306829\work\cpp\build\basix.vcxproj]
D:\bld\fenics-basix-meta_1714546306829\work\cpp\basix\mdspan.hpp(3849): message : see reference to class template instantiation 'std::experimental::detail::static_array_type_for_padded_extent<_PaddingValue,_Extents,_ExtentToPadIdx,_Rank,Enabled>' being compiled [D:\bld\fenics-basix-meta_1714546306829\work\cpp\build\basix.vcxproj]
```

but this went away after upgrading to vs2022. In case anybody in the future reports a similar error, their compiler might be outdated.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -31,6 +31,9 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 option(BUILD_SHARED_LIBS "Build Basix with shared libraries." ON)
 add_feature_info(BUILD_SHARED_LIBS BUILD_SHARED_LIBS "Build Basix with shared libraries.")
 
+option(INSTALL_RUNTIME_DEPENDENCIES "Include runtime dependencies in install (Windows-only)" ON)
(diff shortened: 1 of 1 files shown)


jhale:
Thanks, good news - I think we'll stick with supporting VS2022.

jhale:
Thanks, good news - I think we'll stick with supporting VS2022.

minrk:
Maybe an informative error (or at least warning) if an older version is detected is a good idea?

PR #837, opened 2024-07-01
Update to std::ranges usage

schnellerhase:
Note currently both `std::accumulate` and `std::reduce` which are frequently used in `basix` do not offer `::ranges` equivalents in `C++20`.

### DIFF
--- a/cpp/basix/cell.cpp
+++ b/cpp/basix/cell.cpp
@@ -460,7 +460,7 @@ cell::facet_normals(cell::type cell_type)
   switch (tdim)
   {
   case 1:
-    std::fill(normal.begin(), normal.end(), 1.0);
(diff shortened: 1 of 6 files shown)


garth-wells (approved):

jhale:
I'd be interested to see where we are with C++23 support across Basix platforms - enable it and try the ranges::fold_left?

schnellerhase:
> I'd be interested to see where we are with C++23 support across Basix platforms - enable it and try the ranges::fold_left?

On this PR or another one maybe?

garth-wells:
> > I'd be interested to see where we are with C++23 support across Basix platforms - enable it and try the ranges::fold_left?
> 
> On this PR or another one maybe?

New PR so that this PR doesn’t get held up.

PR #838, opened 2024-07-02
C++23 support

schnellerhase:


schnellerhase:
Hefty changes necessary to adapt to mdspan interface change from `(...)` to `[...]`.

jhale:
Let's assess this now before wasting time moving forward on a patch we cannot merge.

schnellerhase:
Sounds good, so the local build with a GCC version 13.2 and the macos CI build seem to be passing. Besides that it could become tricky to address the absence of a `mdpsan` implementation with c++23 ready compilers. Even trickier is the case of nof c++23 support and neither `mdspan` - i.e. we can probably not make us of `[...]` in these cases.

jhale:
Sounds promising - GCC 11.4.1 on RedHat 9-type distributions is going to be the lower bound.

Regarding midspan, can't we keep on using the Kokkos one in all cases for now?

schnellerhase:
They seem to be all caused by some static assert/requirement. Not sure why this is problem here and otherwise not for now though. 

Will try with the Kokkos `mdspan` - also will check if we are up to date.

jhale:
Kokkos mdspan does claim to support GCC 11 with C++23 with Wall, see their README.

@garth-wells has done the mdspan header copy in the past.

garth-wells:
> Kokkos mdspan does claim to support GCC 11 with C++23 with Wall, see their README.
> 
> @garth-wells has done the mdspan header copy in the past.

See https://github.com/kokkos/mdspan/tree/single-header.

schnellerhase:
OK we have 2 builds passing, macos and the dolfinx integration are working (at least the basix build)

schnellerhase:
Have you guys maybe got an idea what may be the cause for the very weird error I'm getting with this small snippet to test why the ubuntu builds are failing? Note that I can access with a `std::array` as indices the mdspan properly, but when using the `[]` operator the assert fails. Testing with the `pass` function yields the correct size assertion however....

This is observed with gcc version 11.4

```cpp

#include "mdspan.hpp"

template <typename T, std::size_t D>
using mdspan_t = MDSPAN_IMPL_STANDARD_NAMESPACE::mdspan<T, MDSPAN_IMPL_STANDARD_NAMESPACE::dextents<std::size_t, D>>;

template<class...IndexType>
void pass(IndexType... indices)
{
    static_assert(sizeof...(IndexType) == 2);
}

void test()
{
    mdspan_t<double, 2> span;
    static_assert(span.rank() == 2);

    std::array<std::size_t, 2> indices = {0, 0};
    static_assert(indices.size() == 2);
    span[indices];

    pass(0, 0);

    static_assert(std::is_convertible_v<decltype((0,0)), mdspan_t<double, 2>::index_type>);
    static_assert(std::is_nothrow_constructible_v<mdspan_t<double, 2>::index_type, decltype((0,0))>);

    span[0, 0]; // but this fails with (rank() == sizeof...(SizeTypes)) evaluating to False
}

```

schnellerhase:
Okay a small update, I experimented with not updating the `()`  to `[]` operator, be explicitly enforcing the former one. This seems to work, see https://github.com/schnellerhase/basix/pull/1 (up to the intel compiler at least - gotta figure out whats going wrong there, because in principle I think they are c++23 ready). So I suggest we remove the upgrading to the `[]` operator from this PR for now and figure out whats going on there in another later one?

schnellerhase:
Not enforcing the standard fixes the cmake error with the intel compiler.

It looks like they are not fully ready to support the iso standard for c++23 so it gets a different flag - which I assume causes the problem: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-0/std-qstd.html

I am not aware of how I could hint cmake to use this flag, and turning of the standard enforcement also is no good solution I imagine. How should we ago about this?

jhale:
Would be worth checking with a very modern version of Cmake installed in the CI to see if it passes the right flag. Otherwise you can detect the Intel Compiler and pass the flag manually.

schnellerhase:
Newer `cmake` does the trick! I used version `3.30`, are there any preferences on which version we should use in this CI?

cpp/basix/mdspan.hpp
```diff
@@ -1,6 +1,9 @@
 #ifndef _MDSPAN_SINGLE_HEADER_INCLUDE_GUARD_
 #define _MDSPAN_SINGLE_HEADER_INCLUDE_GUARD_
 
+#define MDSPAN_USE_PAREN_OPERATOR 1
```
jhale, line 4: Set these via Cmake with a comment explaining the issue so we will remember to revisit.

.github/workflows/intel.yml
```diff
@@ -24,8 +24,12 @@ jobs:
       CXX: icpx
       DEBIAN_FRONTEND: noninteractive
     steps:
+      - name: Setup cmake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: '3.30.x'
       - name: Install required packages to install Intel compilers and to build
-        run: apt -y update && apt install -y bzip2 cmake git gnupg ninja-build wget
+        run: apt -y update && apt install -y bzip2 git gnupg ninja-build make wget
```
jhale, line 32: Why do you need ninja and make? Should be fine using ninja generator.

.github/workflows/intel.yml
```diff
@@ -24,8 +24,12 @@ jobs:
       CXX: icpx
       DEBIAN_FRONTEND: noninteractive
     steps:
+      - name: Setup cmake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: '3.30.x'
```
jhale, line 30: Did you try anything lower?

schnellerhase: Nope, just bumped it to latest.

.github/workflows/intel.yml
schnellerhase: The build itself went fine, I believe it was the demos build step in the CI that failed if `make` is not available.

jhale: I guess cmake installs make which isn't technically correct.

You can leave it as it is, good to have some test coverage using make.

.github/workflows/intel.yml
jhale: Could you try shifting back? Once known you should set cmake_minimum_version in the CMake file if Intel is detected.

jhale: You can do what you want in the CI (e.g. 3.30) but package maintainers need something precise.

schnellerhase: Should be version $\geq 3.26$ https://gitlab.kitware.com/cmake/cmake/-/commit/4331f7b4bf024c20a06b28664e4f656210de7b5d

jhale: Great. So set CMake minimum version to 3.26 only when Intel is detected.

demo/cpp/demo_create_and_tabulate/CMakeLists.txt
```diff
@@ -1,10 +1,14 @@
 cmake_minimum_required(VERSION 3.16)
 project(demo_create_and_tabulate LANGUAGES CXX)
 
-set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# TODO: remove after issue https://github.com/FEniCS/basix/issues/842 is resolved
+add_compile_definitions(MDSPAN_USE_BRACKET_OPERATOR=0)
```
jhale, line 9: This is old school - target_compile_definitions is the modern CMake way.

python/CMakeLists.txt
```diff
@@ -8,10 +8,21 @@ endif()
 project(basix_nanobind VERSION "0.9.0.0" LANGUAGES CXX)
 
 # Set C++ standard
-set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
+  # c++ 23 of intel not yet iso ready, special flag not set in all cmake versions
```
jhale, line 16: Simpler: Only newer CMake passes non-standard Intel C++23 flag.

jhale:
As an aside, MSVC also doesn't support [P2128R6 Multidimensional subscript operator](https://wg21.link/p2128r6) yet so this is the right way to go for now.

jhale:
And [PR2128R6](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2128r6.pdf) was only added in GCC 12 according to this RedHat article https://developers.redhat.com/articles/2022/04/25/new-c-features-gcc-12#c__23_features

schnellerhase:
> And [PR2128R6](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2128r6.pdf) was only added in GCC 12 according to this RedHat article https://developers.redhat.com/articles/2022/04/25/new-c-features-gcc-12#c__23_features

That could explain my findings with the code snippet before - not sure how it explains the assertion going off though.

jhale (approved):

schnellerhase:
I the website failure caused by the here introduced changes or does this originate from somewhere else?

schnellerhase:
Yeah, very weird, I'm not even sure it's the deprecation warning though. Never used any of the packages that get used for the website construction, so I'm not sure how to go about it.

schnellerhase:
Should be fixed with https://github.com/FEniCS/web/pull/187.

PR #839, opened 2024-07-02
Enum to string

ampdes:
This is only done for the lagrange variant at the moment.


francesco-ballarin:
Similar string to enum functions used to be available in basix, but were removed in #832 by @garth-wells.

Please give us more context on why you need those conversion functions to be reinstated, otherwise it's unlikely we'll decide to (basically) revert #832 and accept this PR.

ampdes:
If this is really necessary is too early to speak about.
This is needed in adios2 based checkpointing to store the enum types as strings.

I will then try to use maps, as is currently done in the C++ demo.

garth-wells:
> If this is really necessary is too early to speak about. This is needed in adios2 based checkpointing to store the enum types as strings.
> 
> I will then try to use maps, as is currently done in the C++ demo.

If something is needed only in the implementation, put it in an anonymous namespace, make private, etc.

garth-wells:
Closing for now.

PR #844, opened 2024-07-08
Switch back to C++20 and use modern set_compile_features to set minimum C++ standard

spraetor:
### Summary
When building `basix` as a a library to be used by other projects, it needs to communicate its c++ standard requirements, especially now since c++23 is required to some extend. Currently the library is configured by setting the global cmake variable `CMAKE_CXX_STANDARD` etc. This works perfectly fine for this library, but is not included in the exported targets. With the current PR the way of enforcing the c++ standard is changed to a compile feature `cxx_std_23`  set directly on the `basix` target. This way it is included in the exported target.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -18,13 +18,6 @@ endif()
 
 include(FeatureSummary)
 
-if(CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
-  # Only newer CMake passes non-standard Intel C++23 flag.
-  # see https://github.com/FEniCS/basix/pull/838
-  # When Intel C++23 support meets ISO standard this can be removed.
-  cmake_minimum_required(3.26) 
-endif()
-
 # Options
 option(BUILD_SHARED_LIBS "Build Basix with shared libraries." ON)
 add_feature_info(BUILD_SHARED_LIBS BUILD_SHARED_LIBS "Build Basix with shared libraries.")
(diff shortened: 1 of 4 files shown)


garth-wells:
We “dropped” some std ranges code in to test. Which of what was added is C++23? 

I’m not sure we want to move to C++23 yet.

jhale:
Thanks @spraetor I didn't know this was possible - we should make this change after discussing whether to use C++23 or C++20 mode.

cpp/CMakeLists.txt
```diff
@@ -114,8 +114,8 @@ target_include_directories(basix PUBLIC
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR};${CMAKE_CURRENT_SOURCE_DIR}>")
 
 # TODO: remove after issue https://github.com/FEniCS/basix/issues/842 is resolved
-target_compile_definitions(basix PRIVATE MDSPAN_USE_PAREN_OPERATOR=1)
-target_compile_definitions(basix PRIVATE MDSPAN_USE_BRACKET_OPERATOR=0)
+target_compile_definitions(basix PUBLIC MDSPAN_USE_PAREN_OPERATOR=1)
+target_compile_definitions(basix PUBLIC MDSPAN_USE_BRACKET_OPERATOR=0)
```
spraetor, line 118: Not sure whether you want to deactivate the `operator[](...)` in the public interface. This might cause problems in user-code.

PR #847, opened 2024-09-02
Expose subentity types to python for easy creation of elements on entities.

jorgensd:
Makes life alot easier when trying to determine what cell types an sub-entity can have (to create mesh independent mixed dimensional forms in DOLFINx).


mscroggs (approved):
Should probably open an issue to wrap all other functions in basix::cell that aren't already wrapped too

PR #853, opened 2024-10-08
Fix return type of sobolev-space in quadrature element

jorgensd:
Needed in exotic code

### DIFF
--- a/python/basix/ufl.py
+++ b/python/basix/ufl.py
@@ -1933,7 +1933,7 @@ class _RealElement(_ElementBase):
     @property
     def basix_sobolev_space(self):
         """Underlying Sobolev space."""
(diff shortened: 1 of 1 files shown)


mscroggs:
The failing DOLFINx tests seems unrelated to this PR?

jorgensd:
> The failing DOLFINx tests seems unrelated to this PR? (but I've removed this from the merge queue just in case)

That is another issue (I've put the test function first, when it should be second.. give me a second to fix it)

mscroggs:
> > The failing DOLFINx tests seems unrelated to this PR? (but I've removed this from the merge queue just in case)
> 
> That is another issue (I've put the test function first, when it should be second.. give me a second to fix it)

https://github.com/FEniCS/dolfinx/pull/3458

PR #872, opened 2024-11-11
Build combined Windows CI run with LLVM/clang-cl

jhale:
This increases Windows CI coverage of Basix by building the combined build with `clang-cl.exe`, a light wrapper around LLVM. This is trivial to implement via vcpkg's host functionality. 

The split build which is relevant to Conda remains on `cl.exe` the Microsoft C/C++ compiler.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -43,8 +43,14 @@ jobs:
           python-version: "3.11"
 
       - name: Install Basix (combined)
-        run: |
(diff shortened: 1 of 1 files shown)


jhale:
The vcpkg package triplets I wanted to use seem unmaintained so this seems like a risk. Perhaps something to revisit in the future.

https://github.com/Neumann-A/my-vcpkg-triplets/tree/master/x64-win-llvm

PR #875, opened 2024-11-15
Fix tolerances for determining DOF transformation type and add test

### LINKED ISSUE #874: Incorrect `dof_transformations_are_permutations` on P5 hex for `float32`
```
import basix
import numpy as np

n = 5
el32 = basix.create_element(basix.ElementFamily.P, basix.CellType.hexahedron, n, basix.LagrangeVariant.gll_warped, dtype=np.float32)
el64 = basix.create_element(basix.ElementFamily.P, basix.CellType.hexahedron, n, basix.LagrangeVariant.gll_warped, dtype=np.float64)

print(el32.dof_transformations_are_permutations, el64.dof_transformations_are_permutations)
```

returns

`False True`

I guess this is due to precision round-off error when trying to determine things automatically. Is this the best/only way?
Anyway, it would be good to fix, and add to tests.

garth-wells:
The check for determining the DOF transformation type was failing for float32 at higher-orders for Lagrange elements. This PR relaxes a tolerance. It is somewhat arbitrary, but does seem to work.

Also makes more use of the BLAS GEMM wrappers to dramatically improve performance at high orders.

Resolves #874.

### DIFF
--- a/cpp/basix/math.h
+++ b/cpp/basix/math.h
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include <algorithm>
 #include <array>
 #include <cmath>
 #include <concepts>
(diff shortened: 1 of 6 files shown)


test/test_lagrange.py
```diff
@@ -949,6 +949,19 @@ def test_dof_transformations_tetrahedron(degree):
         assert np.allclose(t, actual)
 
 
+@pytest.mark.parametrize(
+    "celltype",
+    [basix.CellType.interval, basix.CellType.triangle, basix.CellType.tetrahedron],
```
chrisrichardson, line 954: Hex too?

cpp/basix/math.h
```diff
@@ -179,9 +179,9 @@ std::pair<std::vector<T>, std::vector<T>> eigh(std::span<const T> A,
 }
 
 /// @brief Solve A X = B.
-/// @param[in] A The matrix
-/// @param[in] B Right-hand side matrix/vector
-/// @return A^{-1} B
+/// @param[in] A The matrix.
+/// @param[in] B Right-hand side matrix/vector.@brief
```
mscroggs, line 183:
```suggestion
/// @param[in] B Right-hand side matrix/vector.
```

mscroggs (approved):

PR #882, opened 2024-11-19
Update `basix.ufl.mixed_element` factory function for case of a single element

garth-wells:
With this change, a call to `basix.ufl.mixed_element` with a list of just one finite element simply returns the single element.

This allows an element to be 'constructed' from a list of one element without making a single element into a mixed element (when it's not a mixed element).

### DIFF
--- a/python/basix/ufl.py
+++ b/python/basix/ufl.py
@@ -1985,7 +1985,7 @@ def _compute_signature(element: _basix.finite_element.FiniteElement) -> str:
     return signature
 
 
-@functools.singledisptach
+@functools.singledispatch
 def element(
     family: _typing.Union[_basix.ElementFamily, str],
     cell: _typing.Union[_basix.CellType, str],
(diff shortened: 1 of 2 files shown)


RemDelaporteMathurin (commented):
Thanks for looking into this @garth-wells 

FWIW I agree with @nate-sime in the way that we would like a "mixed element" of only one element to behave like an actual `MixedElement`. The problem isn't only at the construction level but also later on when used.

For example I want to be able to call `ufl.split` on a function of a functionspace built with that "mixed" element. I believe the proposed interface would not make this possible, which means I need to have different ways to handle things depending on the number of sub-elements, which defeats the purpose imo.

RemDelaporteMathurin:
> What functionality is needed (preferably with concrete examples)?

Sure, I tried to be more specific in this comment https://github.com/festim-dev/FESTIM/pull/922#issuecomment-2486792373 although this is hard to parse because it's FESTIM-specific. I'll try to make a pure dolfinx example to show how we use this

PR #885, opened 2024-11-25
Use default value shape for symmetric elements

michalhabera:
An attempt to solve https://github.com/FEniCS/dolfinx/issues/3516


garth-wells (approved):

garth-wells:
Nice to remove a special case.

PR #892, opened 2024-12-14
Encompassing DeRham element

jmv2009:
Attempting to make a generic conformal DeRham element here.
Element: Scope: line, triangle, quadrilateral, tetrahedron, pyramid, prism, hexahedron
Full spaces (P) or trimmed spaces (P-)
Serendipity or tensor product 
Tensor product: Tiniest or Full
2D: Grad Curl or Curl Div

-1 Forms: "Real"
0: Forms: scalar fields
1: Vector fields
..

Following and extending Cockburn & Fu (compatible, admissible)

### DIFF
--- a/cpp/basix/e-DeRham
+++ b/cpp/basix/e-DeRham
@@ -1,0 +1,31 @@
+// Copyright (c) 2024 Johannes M. Vogels
+// FEniCS Project
+// SPDX-License-Identifier:    MIT
+
+#pragma once
+
+#include "cell.h"
+#include "element-families.h"
+#include "finite-element.h"
+#include <concepts>
+
+namespace basix::element
+{
+/// Create a finite element on cell with given degree
+/// @param[in] celltype The cell type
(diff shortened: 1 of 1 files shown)


cpp/basix/e-DeRham.h
```diff
@@ -0,0 +1,31 @@
...
+/// @param[in] rotated2D : Curl Div instead of Grad Curl complex for 2D
+/// @return A finite element
+template <std::floating_point T>
+FiniteElement<T> create_DeRham_element(cell::type celltype,
+                                       int degree,
+                                       int form,
+                                       bool trimmed,
+                                       bool tensor,
+                                       bool tiniesttensor,
+                                       bool rotated2D
```
mscroggs, line 30: I'd strongly favour using an enum here rather than a long list of bools. Would it be helpful in the short term to add a Python utility function to create elements in this way? That could be added very quickly if so

PR #895, opened 2025-01-09
Rational functions for pyramid

jmv2009:
This commit generates the proper orthonormal rational functions for the pyramid.

THE (z) DERIVATIVES MAY NOT BE CORRECT.


jmv2009:
I am changing to pyramid such that there are four slanted phases as in https://www.math.u-bordeaux.fr/~durufle/montjoie/pyramid.php. I also changed the orthogonal rational functions to include an extra factor of 1/(1-z) overall. This makes sure that the first order derivatives of the lower order rational functions are included in the function space as required for the generation of for order complexes of elements c.f. Cockburn and Fu. (to be verified again). For complexes with higher derivatives, more such factor may need to be included. There are still a couple of problems:
a) basix.create_lattice is still on the original pyramid
b) picture?
c) Elements with vertices don't work yet. This will require removing the 1/(1-z) factor, but the orthogonal polynomials will stay the same.

PR #900, opened 2025-02-12
OneAPI -Wall too strict, disable one flag

jhale:
https://github.com/FEniCS/dolfinx/actions/runs/13265704835/job/37033476903?pr=3626#step:7:101


jhale:
This seems to fix the issue. If there are ways to tighten the application of the flag then I'm open to suggestions.

https://github.com/FEniCS/dolfinx/pull/3628

PR #913, opened 2025-03-26
Fix bug in `permute_subentity_closure` for higher order finite elements

### LINKED ISSUE #914: `subentity_permutations` is wrong for higher order elements
Visible through creating a submesh of Lagrange of order >3:
```python
from mpi4py import MPI
import dolfinx
import basix.ufl
import ufl
import numpy as np
mesh = dolfinx.mesh.create_unit_cube(MPI.COMM_WORLD, 2,5,4)
fam = "Lagrange"
deg = 4
V = dolfinx.fem.functionspace(mesh, (fam, deg))

x = V.tabulate_dof_coordinates()
cells = V.dofmap.list

c_el = basix.ufl.element(fam, mesh.basix_cell(), deg, shape=(3, ))

new_mesh = dolfinx.mesh.create_mesh(MPI.COMM_WORLD, cells, x, ufl.Mesh(c_el))

def marker_left(x):
    return np.isclose(x[0], 0.0)

tdim = mesh.topology.dim
fdim = tdim - 1
new_mesh.topology.create_connectivity(fdim, tdim)
left_fac = dolfinx.mesh.locate_entities(new_mesh, tdim - 1, marker_left)
submesh, entity_map, vertex_map, geom_map = dolfinx.mesh.create_submesh(new_mesh, 2, left_fac)

with dolfinx.io.VTXWriter(MPI.COMM_WORLD, "test.bp", submesh) as bp:
    bp.write(0.0)
```

Image

If one tesselate the mesh in Paraview, you will observe that the interior of all the surface cells is not rotated properly (only the ones in the top right corner).
This results in 
```python
print(dolfinx.fem.assemble_scalar(dolfinx.fem.form(1*ufl.dx(domain=submesh))))
```
returning `1.2280701788538921` not `1`

mscroggs:
Resolves #914, i.e. it now performs the correct permutation on faces of higher order function spaces.
Additionally, it exposes the higher-level API of `permute_subentity_closure`/`permute_subentity_closure_inv` in Python, allowing for conforming 2D->1D interpolation between compatible function spaces created using `dolfinx.mesh.create_submesh`. By exposing the higher level API, one avoid having to copy the logic for convert `cell_info` to to `entity_info` into a Python function.

Removes the wrapping of the span as a return function in Python, as the array is modified in-place, aligning with the other permutation functions.


python/wrapper.cpp
```diff
@@ -106,6 +106,15 @@ void declare_float(nb::module_& m, std::string type)
              self.permute_subentity_closure(_d, entity_info, entity_type);
              return as_nbarray(std::move(_d));
            })
+      .def("permute_subentity_closure",
+           [](const FiniteElement<T>& self,
+              nb::ndarray<std::int32_t, nb::ndim<1>, nb::c_contig> d,
```
garth-wells, line 111: Should this be const, i.e.
```cpp
nb::ndarray<const std::int32_t, nb::ndim<1>, nb::c_contig> d,
```

python/wrapper.cpp
```diff
@@ -106,6 +106,15 @@ void declare_float(nb::module_& m, std::string type)
...
+              std::uint32_t cell_info, cell::type entity_type, int entity_index)
+           {
+             std::span<std::int32_t> _d(d.data(), d.shape(0));
+             self.permute_subentity_closure(_d, cell_info, entity_type, entity_index);
+             return as_nbarray(std::move(_d));
```
garth-wells, line 116: It looks wrong to move a `std::span`.

The code changes `d` in place, and the function also returns. It would be cleaner to not change `d` in place and to just return a permuted array

jorgensd (approved):
I made changes to the title and description of the PR, as well as updating the nanobind interface to not returned a wrapped version of the view into the permutation data.

python/wrapper.cpp
jorgensd: No, it follows the same logic as the other permute functions, which does the permutation in-place.

jorgensd:
@mscroggs @garth-wells can we merge this?

PR #926, opened 2025-05-06
Expose gauss jacobi quadrature with alpha parameter

### LINKED ISSUE #922: expose Gauss-Jacobi rules
with parameter a != 0

mscroggs:
Resolves #922


samuel0007:
For my application, I only need an exposed alpha and defined on [0,1]  👍 

But I think from what I understand, one should have (for general alpha and beta) - as the weight function impacts when you pullback the integral:

$$
\zeta_{i}
= \frac{1 + \widehat\zeta_{i}^{(\alpha,\beta)}}{2},
\qquad
\omega_{i}
= \frac{\widehat\omega_{i}^{(\alpha,\beta)}}{2^{\alpha+\beta+1}}.
$$

Which also match this small example I found online [https://www.math.umd.edu/~petersd/460/html/gaussjacobi_ex.html](url)

Also maybe you might also want to use this rescaled function in the `make_quadrature_tetrahedron_collapsed` and `make_quadrature_triangle_collapsed` to make it consistent?

cpp/basix/quadrature.h
```diff
@@ -48,12 +58,14 @@ quadrature::type get_default_rule(cell::type celltype, int m);
...
 /// @brief Get Gauss-Lobatto-Legendre (GLL) points on the interval [0,
 /// 1].
+/// @tparam The floating point type.
 /// @param[in] m Number of points.
 /// @return Array of GLL points.
 template <std::floating_point T>
 std::vector<T> get_gll_points(int m);
 
 /// @brief Get Gauss-Legendre (GL) points on the interval [0, 1].
+/// @tparam The floating point type.
```
chrisrichardson, line 68:
```suggestion
/// @tparam T The floating point type.
```

cpp/basix/quadrature.cpp
```diff
@@ -4976,4 +4989,11 @@ template std::vector<double> quadrature::get_gl_points(int);
 
 template std::vector<float> quadrature::get_gll_points(int);
 template std::vector<double> quadrature::get_gll_points(int);
+
+/// @cond DOXYGEN_SHOULD_SKIP_THIS
```
chrisrichardson, line 4993: I think this should go around *all* the explicit instantiations, e.g. if you look at `get_gll_points` in the generated docs, you will only see `double` at the moment, not `T`, because that was the last thing doxygen saw.

chrisrichardson: I guess there may be other explicit instantiations in other files... worth checking (maybe not in this PR).

chrisrichardson (approved):

cpp/basix/quadrature.cpp
mscroggs: I've opened #929 to remind me to do this

PR #934, opened 2025-05-23
Use quadrature rules from quadraturerules library

mscroggs:


jhale:
My two cents; it would be better to have your library output a header-only C++ version and bundle it into Basix.

PR #952, opened 2025-08-25
allow stable abi builds

minrk:
same as https://github.com/FEniCS/dolfinx/pull/3875

doesn't enable this by default or anything, or even guarantee its success. This only honors scikit-build-core's stable abi inputs so builders can request it.


jhale:
I made some modifications to test using cibuildwheel. Seems to work, and automatically runs tests using cp312-built abi3 wheel on cp313 interpreter.

PR #954, opened 2025-09-03
Drop Python 3.9 support

schnellerhase:
Python 3.9 reaches EoL at 2025-10.


mscroggs:
I've removed 3.9 from the requirements for merging (and added 3.12 as required as it wasn't already). Also I opened #955 to remind us to add 3.13 testing at some point

PR #969, opened 2025-10-21
Use Python version of library as version

jorgensd:
To be able to distinguish between dev and normal releases, we should use the Python version of the package rather than the C++ version, as the Python-versioning has a `.dev0` at the end to indicate pre-release/development release, while this doesn't really exist in the C++ versioning.


jorgensd:
Ref: https://github.com/FEniCS/dolfinx/pull/3960

Issue #993, opened 2026-02-24
Extend mypy coverage to Python demos


schnellerhase:
Currently the Python demos are not covered with the `mypy` check, this should be included in the type checks.

risa-hirsh:
Hi, is this something I could work on? I don't see this labeled as "good first issue" and I'm new to contributing, would this still be appropriate for me to help with?

schnellerhase:
Always, thanks! The checks in the CI can be found here https://github.com/FEniCS/basix/blob/main/.github/workflows/pythonapp.yml#L38-L42.

PR #1003, opened 2026-04-29
Include python demos in mypy check

risa-hirsh:
This pull request is for Issue #993 and extends mypy checks to cover the Python demo files and fixes type issues in those demos.

I made the following changes:
-Added demo/python to the mypy check
-Used typing.cast where necessary to handle ArrayLike return types
-Added minimal NumPy typing annotations to keep everything simple and readable

I didn’t make any changes to the logic of the demos. I compared the results from before and after making changes for every demo to ensure the results did not change. 

I tested that mypy ran on demo/python successfully

### DIFF
--- a/demo/python/demo_custom_element.py
+++ b/demo/python/demo_custom_element.py
@@ -12,6 +12,18 @@ import numpy as np
 import basix
 from basix import CellType, LatticeType, MapType, PolynomialType, PolysetType, SobolevSpace
 
+# Imporrts for type checking
+
+import typing
+
+import numpy.typing as npt
+
+# Aliases for type casting to maintain readability
+
+FloatArray = npt.NDArray[np.float64]
+FloatingArray = npt.NDArray[np.floating]
+QuadratureRule = tuple[FloatArray, FloatArray]
+
 # Lagrange element with bubble
 # ============================
 #
@@ -76,11 +88,14 @@ wcoeffs[3, 4] = 1
 # the largest degree that the integrand will be, so these integrals will
 # be exact).
 
-pts, wts = basix.make_quadrature(CellType.quadrilateral, 4)
-poly = basix.tabulate_polynomials(PolynomialType.legendre, CellType.quadrilateral, 2, pts)
-x = pts[:, 0]
-y = pts[:, 1]
-f = x * (1 - x) * y * (1 - y)
+pts, wts = typing.cast(QuadratureRule, basix.make_quadrature(CellType.quadrilateral, 4))
+poly = typing.cast(
+    FloatArray,
+    basix.tabulate_polynomials(PolynomialType.legendre, CellType.quadrilateral, 2, pts),
+)
+x_coord = pts[:, 0]
+y_coord = pts[:, 1]
+f = x_coord * (1 - x_coord) * y_coord * (1 - y_coord)
 for i in range(9):
     wcoeffs[4, i] = sum(f * poly[i, :] * wts)
 
@@ -100,7 +115,7 @@ for i in range(9):
 #
 # The shape of each of the point lists is (number of points, dimension).
 
-x = [[], [], [], []]
+x: list[list[FloatingArray]] = [[], [], [], []]
 x[0].append(np.array([[0.0, 0.0]]))
 x[0].append(np.array([[1.0, 0.0]]))
 x[0].append(np.array([[0.0, 1.0]]))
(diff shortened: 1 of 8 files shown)


risa-hirsh:
@schnellerhase Hi, I just want to make sure that the failing CI check isn't something I need to worry about. It looks like that same check has been failing on other pull requests recently and isn't related to these changes. Is there anything I need to modify? Thank you!

demo/python/demo_custom_element.py
```diff
@@ -7,11 +7,20 @@
...
 
 import basix
 from basix import CellType, LatticeType, MapType, PolynomialType, PolysetType, SobolevSpace
 
+# Aliases for type casting to maintain readability
+
+FloatArray = npt.NDArray[np.float64]
+FloatingArray = npt.NDArray[np.floating]
+QuadratureRule = tuple[FloatArray, FloatArray]
+
```
schnellerhase, line 23: These are quite verbose changes to every demo for the sake of typing. What is the issue raised without them?

schnellerhase:
> @schnellerhase Hi, I just want to make sure that the failing CI check isn't something I need to worry about. It looks like that same check has been failing on other pull requests recently and isn't related to these changes. Is there anything I need to modify? Thank you!

No mypy checks are broken for unrelated reasons.

demo/python/demo_custom_element.py
risa-hirsh: Thanks for the feedback! The demos had 305 errors with the mypy check which I've attached in a text file: [demo_mypy_errors.txt](https://github.com/user-attachments/files/27253464/demo_mypy_errors.txt)

The majority of errors across all of the demos come from Basix functions returning broad `ArrayLike` types (these functions include `make_quadrature`, `create_lattice`, `tabulate`, `tabulate_polynomials`). Mypy is treating these values as large unions (buffers, strings, nested sequences, etc.), which leads to errors when the demos use them as NumPy arrays, for example:
- slicing/indexing (`pts[:, 0]`, `poly[i, :]`)
- arithmetic (`f * poly * wts`)
- passing them to functions expecting `NDArray`

demo_custom_element.py had the most errors (199) which also included:
1. Empty nested lists such as `x = [[], [], [], []]` and `M = [[], [], [], []]` are inferred as `list[list[Never]]`, so mypy rejects later `.append(np.array(...))` calls and the later `create_custom_element(..., x, M, ...)` calls
2. There are also a few smaller API typing mismatches (using lists where tuples are expected).

The lines you commented on specifically are not necessary, I added them thinking it would keep things more readable but inline type casting would work just fine!

schnellerhase: Right thanks. These are however exactly the errors we would like to resolve with this extended check, so ignoring them by casting to the expected types is not the desired resolution strategy. They show that the current type hints are insufficient. Rather we want to ensure that the type hints provided align with the use cases in the demos. For an example to resolve the errors related to `tabulate_polynomials` and `make_quadrature` see https://github.com/FEniCS/basix/tree/schnellerhase/fix-some-types (feel free to incorporate).

schnellerhase:
@risa-hirsh, the `DOLFINx` related mypy checks should with the merge of https://github.com/FEniCS/basix/pull/1004 now also be resolved. If you update your branch the CI should pass.

mscroggs:
Hi @risa-hirsh, thanks for opening your first pull request to Basix!

I agree with @schnellerhase that it would be good to avoid adding too much typing to the demos. Do you want to have a go at updating the types of the functions as in https://github.com/FEniCS/basix/compare/main...schnellerhase/fix-some-types?

demo/python/demo_custom_element.py
```diff
@@ -156,7 +168,7 @@
 
 element = basix.create_custom_element(
     CellType.quadrilateral,
-    [],
+    (),
```
mscroggs, line 171: This is a correction we should make.

Related: it would be good to update the basix functions to accept any `Sequence` type and convert to tuples inside the functions. I've opened #1010 for this and labelled it good first issue

mscroggs:
Closing this as the issue it was addressing was resolved by #1033.

@risa-hirsh: Thanks for your interest in getting involved in developing Basix. If you'd like to continue getting involved, keep an eye out for [issues tagged good first issue](https://github.com/FEniCS/basix/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22) or reach out to us on the #development channel on the FEniCS Slack.

PR #1004, opened 2026-05-02
Fix: remove PETSc stub for tmp. typing fix

schnellerhase:
Pulls in workaround from https://github.com/fenics-dolfiny/dolfiny/commit/a33092ca17ca752258fd5ad195638b6158938813 for dolfinx not having caught up with new PETSc stubs.

(The underlying issue why this is not a problem in doflinx is that this runs the mypy check without venv on a full blown PETSc aware install - dolfinx is currently missing this)

Ref https://github.com/FEniCS/dolfinx/issues/4059#issuecomment-4287461055.

Depends on https://github.com/FEniCS/dolfinx/pull/4216.

(To make PETSc aware typing checks work here, we need https://github.com/FEniCS/dolfinx/pull/4095)

### DIFF
--- a/.github/workflows/dolfinx-tests.yml
+++ b/.github/workflows/dolfinx-tests.yml
@@ -62,7 +62,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
-          ref: main
+          ref: schnellerhase/mypy-non-venv
       - name: Get DOLFINx
         if: github.event_name == 'workflow_dispatch'
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


.github/workflows/dolfinx-tests.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
-          ref: main
+          ref: schnellerhase/mypy-non-venv
```
schnellerhase:
```suggestion
          ref: main
```

mscroggs:
Do we need to open a housekeeping issue to remind us to remove this later, or is it covered by https://github.com/FEniCS/dolfinx/issues/4059?

schnellerhase:
Good idea, tracked there.

PR #1011, opened 2026-05-15
Fix: cross-compile of wheels

schnellerhase:
Build-wheels is broken on `main` - https://github.com/FEniCS/basix/actions/runs/25927919086. The problem is cross-compilation of the nanobind stubs (in build time mode) - the python architecture of the `nanobind stubgen` invocation and the target wheel architecture do not necessarily align. In this case `nanobind` will fail in install time mode.

Changes to all stubs being generated at install time.

Tested `build-wheels` at https://github.com/FEniCS/basix/actions/runs/25930015175.

Ref https://github.com/FEniCS/basix/pull/992.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -113,42 +113,19 @@ else()
 endif()
 
 # and the nanobind typing stubs.
-if (WIN32)
-  # On Windows we cannot import basix._basixcpp without installing the package
-  # alongside its external dlls.
+# We cannot rely on import without installing the package for cross compilation of wheels.
 
-  # This *must* be called prior to nanobind_add_stub
-  install(TARGETS _basixcpp LIBRARY DESTINATION basix)
-  # nanobind automatically installs into ${CMAKE_INSTALL_PREFIX} in
-  # INSTALL_TIME mode.
-  # DEPENDS is redundant in INSTALL_TIME mode - _basixcpp must be installed.
-  nanobind_add_stub(
-    _basixcpp_stub
-    MODULE basix._basixcpp
-    INSTALL_TIME
-    MARKER_FILE basix/py.typed
-    VERBOSE
-    OUTPUT
-      basix/_basixcpp.pyi
-  )
-else()
-  # On UNIX-like systems we can import the cpp compiled module before
-  # installing and rely on dynamic linking at import time.
-  nanobind_add_stub(
-    _basixcpp_stub
-    MODULE _basixcpp
-    PYTHON_PATH $<TARGET_FILE_DIR:_basixcpp>
-    DEPENDS _basixcpp
-    MARKER_FILE ${CMAKE_CURRENT_SOURCE_DIR}/basix/py.typed
-    VERBOSE
-    OUTPUT
-      _basixcpp.pyi
-  )
-
-  install(TARGETS _basixcpp LIBRARY DESTINATION basix)
-  install(FILES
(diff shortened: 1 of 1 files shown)


jhale:
Isn't this a bug in nanobind?

schnellerhase:
Isn't the problem that reading a binary, which was built for a different platform will never work? Why would that be a bug?

jhale:
No, macOS 14 ARM has Rosetta 2 translation and can run x86 wheels.

https://cibuildwheel.pypa.io/en/stable/platforms/#macos-architectures

schnellerhase:
That would require the python interpreter to be run through rosetta, which is not happening here.
> ... on Apple Silicon when running Python under Rosetta 2 emulation.

My understanding is that the macOS runs are failing here because they are currently the only runs which actually perform a cross compile. Nothing special besides that to the platform I think.

jhale:
That's exactly my point; nanobind's stubgen python interpreter is not being invoked properly, so it's a nanobind bug.

schnellerhase:
What I was pointing at is that there is never a python interpreter being run with rosetta translation in place here for the whole wheel build, so there is no other interpreter for nanobind to pick up on. Same would hold true for any other platform running a cross compile (where a rosetta or alike is not available).

schnellerhase:
See also https://gitlab.com/petsc/petsc/-/merge_requests/8366 where this problem was also encountered and resulted in design constraints for the `PETSc` equivalent.

jhale:
I don't agree; take a look around at https://github.com/pypa/cibuildwheel/blob/main/cibuildwheel/platforms/macos.py#L409 where you can see 'arch' and 'pip' being used.

schnellerhase:
I see, `pypa/cibuildwheel` does not perform (or support) a cross compilation but considers the x86 on arm for macOS native....for whatever reason. And `nanobind`'s stubgen seems to be not supporting cross compilation anyhow if I see correctly.
> For this to work, the module must be importable.

No idea why install time then works any more reliable than build time here, and not the other way around.

jhale:
Should be possible to fix within nanobind:

https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html#variable:CMAKE_OSX_ARCHITECTURES
https://cmake.org/cmake/help/latest/prop_tgt/OSX_ARCHITECTURES.html#prop_tgt:OSX_ARCHITECTURES

with the `MODULE` being passed having the needed metadata.

jhale:
> No idea why install time then works any more reliable than build time here, and not the other way around.

A good point, worth looking into.

jhale:
Just tried this locally using fat-binary macOS python.org install; stubgen and the resulting basix work perfectly!

`arch -x86_64 python -m pip install .`

 Must be an environment/cibuildwheel issue.

python/CMakeLists.txt
```diff
@@ -113,42 +113,15 @@ else()
...
+
+# Note: install time mode required for wheel cross compilation.
+nanobind_add_stub(
+  _basixcpp_stub
+  MODULE basix._basixcpp
+  INSTALL_TIME
+  MARKER_FILE basix/py.typed
+  VERBOSE
+  OUTPUT
+    ${CMAKE_INSTALL_PREFIX}/basix/_basixcpp.pyi
```
schnellerhase, line 126: @jhale I noticed we need the prefix on Unix when testing. Are we sure it works without on Windows?

schnellerhase:
Superseded by https://github.com/FEniCS/basix/pull/1012.

PR #1012, opened 2026-05-18
Move to native macOS wheel builds

jhale:
https://github.com/FEniCS/basix/pull/1011 discovered that building x86-64 wheels via Rosetta emulation on aarch64 runners leads to the nanobind stubgen process failing due to the x86-64 wheel not being loadable. I cannot recreate this locally - `arch -x86_64 pip install .` works perfectly. This means it is probably a `cibuildwheel` or runner configuration issue.

That said, `macos-14` runners will shortly be deprecated and GitHub provides aarch64 `macos-26` and x86-64 `macos-26-intel` runners which bypass this issue.

This PR:
- Bumps minimum Python to 3.11.
- Bumps builders to `macos-26` and `macos-26-intel` and builds/tests native wheels.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -40,7 +40,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, ubuntu-24.04-arm, macos-14, macos-15, windows-2025]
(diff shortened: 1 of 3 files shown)


schnellerhase (commented):
Looks good!

Do we want to drop 3.10 already here? It is not yet EoL.

jhale:
Yes, was discussed the other day with @mscroggs.

PR #1014, opened 2026-05-19
Module level pytest skip for numba

jhale:
If it doesn't need to use numba, it can be elsewhere.

This will allow the majority of tests to run on platforms without numba/llvmlite wheels, for example, x86-64 macOS and ARM Windows. Wheel build tests for non-numba platforms are re-enabled.

I also removed manual calls to `abi3audit` - `cibuildwheel` does this automatically now.


jhale:
Seems like rubygems.org is having some issues leading to package pull timeouts.

PR #1026, opened 2026-06-15
Add `gersemi` CI check

schnellerhase:
Adds a formatting check for `CMake` files.

Same setup as https://github.com/FEniCS/dolfinx/pull/4242.


jhale:
Result looks nice to me.

Issue #1020, opened 2026-06-05
Change numbering of sub-entities?


mscroggs:
Currently, sub-entities of the reference cell are numbered following the [UFC convention](https://fossies.org/linux/ufc/doc/manual/ufc-user-manual.pdf) on simplex cells:

> When the vertices have been numbered, the remaining mesh entities are numbered within each topological dimension based on a lexicographical ordering of the corresponding ordered tuples of non-incident vertices.

However non-simplex cells instead are ordered by **incident** vertices. This matches the convention used by [DefElement](https://doi.org/10.1007/s44207-026-00011-0):

>  <img width="761" height="314" alt="Image" src="https://github.com/user-attachments/assets/32928718-e712-40d0-b0e8-0de77a47c8d0" />

This inconsistency is a little confusing. I propose that we update the numbering of sub-entities on simplex cells so that all cells match the DefElement conventions

PR #1028, opened 2026-06-17
Renumber simplex reference cells

mscroggs:
Renumber the entities of the triangle and tetrahedron reference cells to follow the consistent convention proposed in #1020.

Following this PR. The triangle and tetrahedron reference cells are numbered like this:

image

image

For reference, the numbering before this PR is:

image

image


jorgensd:
This has to be introduced very carefully, as this will break all IO permutations we have made in DOLFINx and any other IO library.

jorgensd:
> This has to be introduced very carefully, as this will break all IO permutations we have made in DOLFINx and any other IO library.


Furthermore, this will break any checkpoint made by adios4dolfinx or io4dolfinx made prior to this merge, as this will affect the element dof layout and I guess the  cell permutations and facet permutation will differ?

mscroggs:
> This has to be introduced very carefully, as this will break all IO permutations we have made in DOLFINx and any other IO library.

Yes, hence "draft" for now

PR #1032, opened 2026-06-30
Fix up `gersemi` check

schnellerhase:
Reverts explicit file listing in `gersemi` linting check of https://github.com/FEniCS/basix/pull/1026.


jhale:
It's strange as I certainly didn't add these for fun :). I genuinely remember getting more fails when I added explicit files.

I will check again locally.

PR #1033, opened 2026-07-04
Extend typing checks to demos

### LINKED ISSUE #993: Extend mypy coverage to Python demos

### LINKED ISSUE #1010: Allow element creation functions to accept any `Sequence` type for value shape
Currently value shape in element creation functions (eg [create_custom_element](https://github.com/FEniCS/basix/blob/37db4b1c85b917682ee09a9d8d9c84e2f6cbdfae/python/basix/finite_element.py#L632)) is typed to be `tuple[int, ...]`.

It would be better for the type hint to be `Sequence[int]` and to convert to a tuple inside the function before calling the wrapped C++ functions

schnellerhase:
Basix is now fully `mypy` covered. Introduces minimal typing generics to `FiniteElement` to pass checks, this should be extended on in the future. 

- Fixes https://github.com/FEniCS/basix/issues/993.
- Fixes https://github.com/FEniCS/basix/issues/1010.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -164,8 +164,8 @@ jobs:
         run: mypy -p basix
       - name: Run mypy (test)
         run: mypy test
-      # - name: Run mypy (demo)
-      #   run: mypy demo/python
+      - name: Run mypy (demo)
+        run: mypy demo/python
 
   build-separately:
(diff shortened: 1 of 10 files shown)


jorgensd (approved):
LGTM

PR #1037, opened 2026-07-09
More typing

schnellerhase:
Caught on dolfinx's demos https://github.com/FEniCS/dolfinx/pull/4280.


.github/workflows/fenicsx-refs.env
```diff
@@ -3,4 +3,4 @@ ufl_ref=main
 ffcx_repository=FEniCS/ffcx
 ffcx_ref=main
 dolfinx_repository=FEniCS/dolfinx
-dolfinx_ref=main
+dolfinx_ref=schnellerhase/mypy-demo
```
schnellerhase:
```suggestion
dolfinx_ref=main
```

PR #1054, opened 2026-08-15
Update mdspan

garth-wells:


jorgensd:
@garth-wells could you update the title of the PR to reflect the various other changes you made in this work as well, as this is what goes into the auto-generated summary on release.

PR #1063, opened 2026-08-20
Enable Nanobind split build

jhale:
This beta initiative in nanobind allows for split mode builds, fully described: https://nanobind.readthedocs.io/en/latest/split_mode.html

Stable API and targeted/optimised API builds are still supported.

The `scikit-build-core` end is not totally smooth, not a big fan of dynamic metadata - I guess this will be improved upstream.

I used Claude Opus 5.0 to sketch out this commit and I agree to the default disclaimer.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -49,6 +49,9 @@ jobs:
 
       - name: Build wheels
         uses: pypa/cibuildwheel@v4.2.0
+        env:
+          BASIX_NANOBIND_SPLIT: "true"
+          CIBW_ENVIRONMENT_PASS_LINUX: BASIX_NANOBIND_SPLIT
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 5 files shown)


jhale:
https://github.com/FEniCS/basix/actions/runs/32352423277 - seems to work well.

jhale:
I propose merging once nanobind 1.0.0 goes final.

jorgensd:
Looks good to me. I think it is ok to keep it optional for now, until we see clear benefits of the split approach.
FYI: @wjakob here is our attempt to use split-build.

jhale:
The feature is aimed at package builders/distributors. There is no benefit of a split build for people building their own Basix - it should always remain optional.

wjakob:
Neat! Nanobind 3.0.0 is out now.

wjakob:
Regarding

> until we see clear benefits of the split approach

You should not see any functional or performance difference. Split mode is designed so that nanobind behaves exactly the same. The only obvious benefit is that the wheel matrix/(tensor?) collapses along one dimension, making the final number of wheels ~5x smaller.

